PR #31, opened 2016-12-11
Fix implementation of napi_create_boolean

gabrielschulhof:
This is a leftover copy/paste error that causes booleans to be rendered as numbers when they return to JS.

### DIFF
--- a/src/node_jsrtapi.cc
+++ b/src/node_jsrtapi.cc
@@ -501,7 +501,7 @@ napi_value napi_create_number(napi_env e, double v) {
 napi_value napi_create_boolean(napi_env e, bool b) {
   JsErrorCode error = JsNoError;
   JsValueRef booleanValue = nullptr;
-  error = JsDoubleToNumber(b, &booleanValue);
+  error = JsBoolToBoolean(b, &booleanValue);
   return reinterpret_cast<napi_value>(booleanValue);
 }
 


digitalinfinity:
LGTM - thanks for fixing! cc @boingoing

mhdawson (approved):
LGTM

mhdawson:
What about the similar fix for the other branches ? We at least need it in the 6.2 branch as I'm thinking we'll use that as the abi-stable-node "master" when we want to rebase onto the latest node.

gabrielschulhof:
@mhdawson this fix is chakracore-specfic, and AFAICT api-prototype-chakracore-7.x is the only branch that contains chakracore, and therefore src/node_jsrtapi.cc. Thus, there should be no other branches to which this fix needs to be applied.

mhdawson:
@gabrielschulhof sorry should have looked more closely at what file was being changed.

gabrielschulhof:
@mhdawson NP :) I too have often confused node_jsvmapi.cc with node_jsrtapi.cc.

ianwjhalliday:
I was not happy with how those file names turned out.  Would be great for someone to reorganize or rename them to clearly show they are v8 and jsrt implementations of napi.  `node_napi_v8.cc` and `node_napi_jsrt.cc`?

(drive by holla')

gabrielschulhof:
Do I need to do anything with this PR before it can be merged?

gabrielschulhof:
Closed in e2480ba998861a0b5d174a9ac62b359e711f7d16

PR #32, opened 2016-12-12
Add comment about string encoding.

fhinkel:

JS VM Api

##### Description of change

In V8, strings are `utf16`. Would it make sense to have a `napi_get_string_utf16`? Or is conversion to `utf8` a performance hit we don't mind?


bnoordhuis:
Sounds sensible to me.  ChakraCore, JavaScriptCore and Spidermonkey all seem to support it and at least in Spidermonkey it's the canonical (i.e. fast) representation.

mhdawson (approved):
LGTM seems reasonable to me as well.

PR #33, opened 2016-12-12
Add comment about own properties

fhinkel:

JS VM Api

##### Description of change

Not V8 related, but do we need methods for own properties as well? Also, do we need a method that returns the property without invoking getters and interceptors (similar to v8::GetRealNamedProperty)?


bnoordhuis:
Engine support may be a bit sketchy.  JavaScriptCore to the best of my knowledge doesn't support anything like it and some quick grepping suggests ChakraCore doesn't either.

EDIT: Doesn't mean adding a TODO comment is a bad idea.  You have my LGTM if you want it but can you drop the whitespace change?

mhdawson:
Opening an issue for discussion as an alternative to adding the TODO comment is also an option.

ianwjhalliday:
Unless performance critical, this sort of thing is usually accomplished by calling the JavaScript built-in.  I.e. `Object#hasOwnProperty` in this case.

aruneshchandra:
Added issue #67 to track this. Closing this PR.

PR #34, opened 2016-12-12
Remove comment regarding releasing weakrefs

fhinkel:

JS VM Api

##### Description of change

I think V8 can't distinguish between `persistent` and `weakref`, so the method to release weakrefs is needed. 

`value` and `persistent` is about lifetime, not stack vs heap, right? If `napi_value` can be stored on the heap, we need to careful in V8 to handle this correctly. Is the section about "Methods to control object lifespan" pretty much in its final form?

### DIFF
--- a/src/node_jsvmapi.h
+++ b/src/node_jsvmapi.h
@@ -260,7 +260,6 @@ NODE_EXTERN napi_value napi_get_persistent_value(napi_env e, napi_persistent p);
 NODE_EXTERN napi_weakref napi_create_weakref(napi_env e, napi_value v);
 NODE_EXTERN bool napi_get_weakref_value(napi_env e, napi_weakref w,
                                         napi_value *pv);
-// is this actually needed?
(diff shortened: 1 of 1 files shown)


src/node_jsvmapi.h
```diff
@@ -259,8 +259,6 @@ NODE_EXTERN void napi_release_persistent(napi_env e, napi_persistent p);
 NODE_EXTERN napi_value napi_get_persistent_value(napi_env e, napi_persistent p);
 NODE_EXTERN napi_weakref napi_create_weakref(napi_env e, napi_value v);
 NODE_EXTERN bool napi_get_weakref_value(napi_env e, napi_weakref w,
-                                        napi_value *pv);
```
bnoordhuis, line 262: This line should not be removed.

mhdawson (approved):
LGTM

mhdawson:
In terms of it being complete, its what we have needed so far, changes would be driven by input from porting additional modules.

Issue #27, opened 2016-12-05
Correct abstraction for try/catch 


mhdawson:
* **Version**: ALL
* **Platform**: ALL
* **Subsystem**: API

Comment from Ben in : https://github.com/nodejs/node-eps/pull/20#issuecomment-264451567

I think most JS engines have an 'is exception pending?' function; V8 is the outlier with its TryCatch construct. API strawman:

// Wrapper around a v8::TryCatch.  Dummy with other engines.
typedef struct napi_trycatch { struct napi_trycatch *v; } napi_trycatch;
NODE_EXTERN void napi_trycatch_new(napi_env e, napi_trycatch *trycatch);
// Returns exception object or undefined.
NODE_EXTERN napi_value napi_trycatch_exception(napi_env e, napi_trycatch trycatch);
NODE_EXTERN void napi_trycatch_delete(napi_env e, napi_trycatch trycatch);

gabrielschulhof:
The V8 documentation [claims](https://v8docs.nodesource.com/node-7.2/d4/dc6/classv8_1_1_try_catch.html#af28a8ba7e1cb6e66bc2468faa4c86eec) that the `TryCatch` should be stack-allocated, so I'm not sure if we can go with the new/delete paradigm. If we can't, we may have to provide a way of calling a function that also catches exceptions, like maybe with an out parameter:

```C
napi_value
napi_call_function(napi_env e, napi_value scope,
                   napi_value func, int argc, napi_value* argv, napi_value* exception);
```

Then, applications can simply re-throw the exception if our implementation sets it.

OTOH, if we *can* heap-allocate, IINM in the V8 implementation we have to remember to `v8::TryCatch::ReThrow()` the exception in `napi_trycatch_delete()` if `napi_trycatch_exception()` has not been called because V8 automatically removes the exception state if a `TryCatch` is within scope. So, in V8 we need at least one additional piece of data in addition to the `v8::TryCatch` itself: A sticky bit `exception_retrieved` which will be set in `napi_trycatch_exception()` to indicate that we need not call `ReThrow()` in `napi_trycatch_delete()`, because the exception has been retrieved.

This way, we have the `JsGetAndClearException()` semantics in `napi_trycatch_exception()`, but also the need to have an object that bookends the transaction that may throw à la `v8::TryCatch`.

[...]

PR #43, opened 2017-01-08
Try catch chakracore

gabrielschulhof:
Implements the try/catch API mentioned in https://github.com/nodejs/abi-stable-node/issues/27#issue-193559367, first porting test/addons-abi and implementing the Makefile modifications necessary to make them run.


src/node_jsvmapi.h
```diff
@@ -55,6 +55,7 @@ struct napi_module_struct {
   struct node_module* nm_link;
 };
 
+#include <node.h>
```
jasongin, line 58: I don't think this is needed here.

jasongin:
Can you move this PR onto the new **api-prototype-chakracore-8.x** branch? It will greatly simplify the PR, since the new branch restores all the tests cases which had been unintentionally deleted from the chakracore-7.x branch.

Issue #44, opened 2017-01-11
Property accessors


jasongin:
NAPI is currently missing support for property accessors, to enable C++ callbacks for getters and setters of properties on a wrapped object.

NAN supports this via the [`Nan::SetAccessor()`](https://github.com/nodejs/nan/blob/master/nan.h#L1908) function, which eventually calls `v8::Object::SetAccessor()` or `v8::ObjectTemplate::SetAccessor()`. NAPI currently has functions for getting and setting property values, but not for setting accessors.

[From the data I collected](https://github.com/nodejs/abi-stable-node/pull/36/files#diff-22c81010be429d42f1edc36047f7b9b9R90), 5 of the top 12 native packages use one or both of those APIs: canvas, fibers, nodegit, sqlite3, zmq. The canvas module implements accessors for several properties like `canvas.width`, `canvas.height`, etc. The sqlite3 module only has a single getter for `database.open`.

jasongin:
Fixed in the 6.2 branch with https://github.com/nodejs/abi-stable-node/commit/67f4581fd711cbb8bec0858f4c653e86792e4aae

To be ported to other branches soon.

jasongin:
Also done in ChakraCore as 27b3fc708b49f6b755486aa47fab09bc30062aac and merged into the V8 8.x branch as 0bdc1b88eb2e186fc3cd979a910d6d3311022491

PR #45, opened 2017-01-12
Property accessors 6.2.0

jasongin:
See https://github.com/nodejs/abi-stable-node/issues/44

This change adds a new `napi_set_accessor` API which sets getter and setter accessors for a property on an object. In support of that, I refactored some of the callback-related code: the callback info opaque pointer type is now shared between function callbacks and accessor callbacks, so I renamed it from `napi_func_cb_info` to `napi_callback_info`, and renamed a few other things accordingly. In the internal implementation, there is a new `CallbackInfoWrapper` and subclasses that wrap the v8 `FunctionCallbackInfo` and `PropertyCallbackInfo` classes.

I assume we're not too concerned about breaking changes at this point, though of course any existing code using NAPI will need to be updated. I am updating the `abi-stable-node-addon-examples` repo for these changes, along with new examples for using property accessors.

### DIFF
--- a/src/node_jsvmapi.cc
+++ b/src/node_jsvmapi.cc
@@ -1,4 +1,4 @@
-/*******************************************************************************
+﻿/*******************************************************************************
  * Experimental prototype for demonstrating VM agnostic and ABI stable API
  * for native modules to use instead of using Nan and V8 APIs directly.
  *
@@ -26,8 +26,6 @@
 void napi_module_register(void* mod) {
   node::node_module_register(mod);
 }
-
-typedef void napi_destruct(void* v);
 
 namespace v8impl {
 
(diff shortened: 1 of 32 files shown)


jasongin:
Based on I've learned today I'm aware of at least a couple pieces missing from this PR:

1. `NAPI_GETTER` and `NAPI_SETTER` macros in `node_api_helpers.h`
2. Changes to `napi_create_constructor_for_wrap_with_methods` to also support accessors.

I'm working on these, but meanwhile I'd still appreciate feedback on the rest of the changes.

mhdawson (commented):
I'll need a bit more time to go through this in more detail (sounds like there are more changes to come anyway), but overall it looks reasonable to me.

One specific question sor far is if we can move some of the common functions into CallbackInfoWrapper instead of duplicating them in each type (for example: This(), Holder() , SetReturnValue, etc.)

mhdawson:
On the breaking changes front, agreed we still need to make breaking changes and as you say we need to then update our existing module ports and the add on examples so they continue to work.  

On that front I'm thinking we might want to set up some ci jobs, I'll think about that a bit more when I get back from vacation.

jasongin:
Regarding `CallbackInfoWrapper`, I did spend some time trying to come up with something less repetitive. It's tricky because v8's `FunctionCallbackInfo` and `PropertyCallbackInfo` have a lot of the same methods but don't have any inheritance relationship. I tried doing something with templates, but it got very complicated to the point that I thought it wasn't worth it.

jasongin:
I pushed another commit that includes the following changes: 
 * Changed `napi_method_descriptor` to `napi_property_descriptor` and added several fields. This is now a C++ equivalent to the property descriptor object passed to the JS `Object.defineProperty()`.
 * Updated `napi_create_constructor()` to take an array of property descriptors (instead of just the method descriptors), so it is able to define methods, accessors, and values for the class.
 * Added `napi_define_property()` that also takes a property descriptor and uses it to define a method, accessors, or value property on an object.
 * Refactored the internal callback wrapper helper classes and functions so they are cleaner and support passing a `void*` data pointer to any function or accessor callback.
 * Added `napi_cb_get_data()` to retrieve the callback data.
 * Added `NAPI_GETTER` and `NAPI_SETTER` macros in `node_api_helpers.h`. They're actually just the same as `NAPI_METHOD` because the args and return value are all handled via the callback info object in the same way. (Unlike nan where `NAN_SETTER` declares the value as a separate callback arg.)

I think it's ready for a full review now.

jasongin:
Oh, I still need to add/update tests covering these changes under `test/addons-abi`.

jasongin:
Pushed another commit with added/updated tests.

gabrielschulhof:
LGTM

jasongin:
Waiting for the [try-catch PR](https://github.com/nodejs/abi-stable-node/pull/49) to be merged, then I will rebase this PR on top of that.

jasongin:
I squashed and rebased on top of the latest changes. ~~but there are some test failures now. Please stand by.~~

jasongin:
OK I fixed the problems after rebasing. This is ready now. Waiting for signoff from @boingoing who said he would review this afternoon.

src/node_jsvmapi.cc
```diff
@@ -316,17 +522,47 @@ napi_value napi_create_constructor_for_wrap_with_methods(
...
+
+      tpl->PrototypeTemplate()->SetAccessor(
+        propertyname,
+        v8impl::GetterCallbackWrapper::Invoke,
+        p->setter ? v8impl::SetterCallbackWrapper::Invoke : nullptr,
+        cbdata,
+        v8::AccessControl::DEFAULT,
+        attributes);
+    }
+    else {
```
boingoing, line 559: Just a nitpick, we could assert p->value here or raise an error condition if it's null. Fine to do this as part of the error handling pass later, though.

src/node_jsvmapi.cc
```diff
@@ -581,63 +864,51 @@ napi_value napi_get_true(napi_env e) {
...
   NAPI_PREAMBLE(e);
-  const v8::FunctionCallbackInfo<v8::Value> *info =
-            v8impl::V8FunctionCallbackInfoFromJsFunctionCallbackInfo(cbinfo);
-  return v8impl::JsValueFromV8LocalValue(info->Holder());
+  v8impl::CallbackWrapper* info =
+    reinterpret_cast<v8impl::CallbackWrapper*>(cbinfo);
+  return info->Holder();
+}
+
+void* napi_get_cb_data(napi_env e, napi_callback_info cbinfo) {
```
boingoing, line 908: Handling of this data pointer is clean throughout. Good stuff.

boingoing:
This looks very good to me. Thanks a lot for doing the work, especially for updating the tests and adding the data pointer to the callback info.

mhdawson (commented):
A bit late but going through the changes look good to me. 

One last question.  Do you think we still need napi_set_property ? The places where it was replaced by napi_define_property seem more concise.

jasongin:
Yes, we still need `napi_set_property`, because sometimes you just want to assign a value to a property (which will invoke any existing set accessor if there is one) without redefining the property.

src/node_jsvmapi.cc
jasongin: I was assuming we should preserve whatever is the existing behavior of PrototypeTemplate()->Set() when it is given a null value.

PR #47, opened 2017-01-15
InstanceOf 6.2.0

jasongin:
Adding a `napi_instanceof` API that checks whether a `napi_value` object is an instance created by a `napi_value` constructor function.

The V8 implementation isn't entirely trivial: V8 code would normally use `FunctionTemplate::HasInstance()`, except here we don't have the `FunctionTemplate` available because (as a V8-only concept) it is not exposed via NAPI. So instead we can implement the instance-of logic in a slightly less efficient way by looking through the prototype chain. (The JSRT implementation can simply call `JsInstanceOf()`.)

I started porting the `canvas` package and found it makes frequent use of `FunctionTemplate::HasInstance()` checks to support sort of polymorphic behavior for drawing different kinds of things on the canvas. The `sqlite3` package use the API in one place, just for a simple argument check. Based on my previous scans, not very many other packages use it.

### DIFF
--- a/src/node_jsvmapi.cc
+++ b/src/node_jsvmapi.cc
@@ -815,6 +815,36 @@ napi_value napi_new_instance(napi_env e, napi_value cons,
   return v8impl::JsValueFromV8LocalValue(result);
 }
 
+NODE_EXTERN bool napi_instanceof(napi_env e, napi_value obj, napi_value cons) {
+  v8::Isolate *isolate = v8impl::V8IsolateFromJsEnv(e);
+  v8::HandleScope scope(isolate);
+
+  v8::Local<v8::Context> context = isolate->GetCurrentContext();
+  v8::Local<v8::Object> v8obj =
+    v8impl::V8LocalValueFromJsValue(obj).As<v8::Object>();
+  v8::Local<v8::Function> v8cons = v8impl::V8LocalFunctionFromJsValue(cons);
+
+  // Can't use v8::FunctionTemplate::HasInstance() because the FunctionTemplate
+  // cannot be obtained from a Function. Instead, recurse up the prototype
+  // chain and check if any prototype's constructor is the specified function.
+
+  v8::Local<v8::String> conskey =
+    v8::String::NewFromUtf8(isolate, "constructor",
+      v8::NewStringType::kInternalized).ToLocalChecked();
+
+  v8::Local<v8::Object> proto = v8obj->GetPrototype().As<v8::Object>();
+  while (!proto->IsNull()) {
+    v8::Local<v8::Function> protocons =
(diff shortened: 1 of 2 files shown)


gabrielschulhof:
I have the V8 implementation.

On Sun, Jan 15, 2017 at 10:43 PM, Jason Ginchereau <notifications@github.com

src/node_jsvmapi.cc
```diff
@@ -815,6 +815,36 @@ napi_value napi_new_instance(napi_env e, napi_value cons,
...
+
+  // Can't use v8::FunctionTemplate::HasInstance() because the FunctionTemplate
+  // cannot be obtained from a Function. Instead, recurse up the prototype
+  // chain and check if any prototype's constructor is the specified function.
+
+  v8::Local<v8::String> conskey =
+    v8::String::NewFromUtf8(isolate, "constructor",
+      v8::NewStringType::kInternalized).ToLocalChecked();
+
+  v8::Local<v8::Object> proto = v8obj->GetPrototype().As<v8::Object>();
```
gabrielschulhof, line 835: It has been my experience that `GetPrototype()` doesn't really work as expected. You can test your implementation by copying https://github.com/nodejs/abi-stable-node/blob/api-prototype-6.2.0/deps/v8/test/mjsunit/instanceof.js and https://github.com/nodejs/abi-stable-node/blob/api-prototype-6.2.0/deps/v8/test/mjsunit/instanceof-2.js and replacing the `instanceof` operators with calls to `napi_instanceof()`.

gabrielschulhof:
@jasongin if you only need instanceof because of canvas and sqlite then perhaps we need not add it to NAPI because it can be implemented using existing NAPI APIs (https://github.com/gabrielschulhof/abi-stable-node/blob/test-instanceof-master/test/addons-abi/test_instanceof/test_instanceof.cc#L15).

If we do add it to NAPI, we should make sure that it passes the test from my [PR](https://github.com/nodejs/abi-stable-node/pull/48).

gabrielschulhof:
Hmmm, given that we currently have canvas, sqlite, and even iotivity-node that need this, maybe we should have an implementation of this in NAPI after all.

jasongin:
I think people would expect it to be part of the API. This PR worked in my limited testing, enough to unblock canvas, but it looks like you've done a lot more extensive testing than I did.

src/node_jsvmapi.cc
jasongin: So what is the difference between `GetPrototype()` and the `__proto__` property? NAPI currently has a `napi_get_prototype()` API that just calls `GetPrototype()`, so is that correct or not?

jasongin:
Closing this in favor of @gabrielschulhof's [more complete PR](https://github.com/nodejs/abi-stable-node/issues/48).

PR #48, opened 2017-01-16
Add instanceof master v8

gabrielschulhof:
Implementation of `napi_instanceof()` which uses the `"prototype"` property of the constructor and the `"__proto__"` property of the object to be examined.

It also adds a test to make sure that the result fully implements the semantics of the `instanceof` operator. The test is taken from V8's https://github.com/nodejs/abi-stable-node/blob/api-prototype-6.2.0/deps/v8/test/mjsunit/instanceof.js and https://github.com/nodejs/abi-stable-node/blob/api-prototype-6.2.0/deps/v8/test/mjsunit/instanceof-2.js and modified to replace the JS `instanceof` operator with calls to the addon that exposes `napi_instanceof()`.

Note that `napi_instanceof()` is not "linearly independent", in that it can be implemented with calls to existing NAPI APIs (https://gist.github.com/gabrielschulhof/3a8d174b860db0c18a732c0b25c12cbb). Unfortunately this currently means that `napi_instanceof()` will be less efficient on V8 than the native V8 `HasInstance()` because we cannot call `FunctionTemplate::HasInstance()`, whereas on ChakraCore `napi_instanceof()` will be more efficient than an implementation using other napi APIs, because we can call `JsInstanceOf()`.

So, adding `napi_instanceof()` will be a net benefit for ChakraCore and a net detriment for V8 - with the benefit that there will be a VM-neutral API for accomplishing this task.

Re #47


test/addons-abi/test_instanceof/test1.js
```diff
@@ -0,0 +1,344 @@
...
+}
+
+function assertEquals(leftHandSide, rightHandSide) {
+	assert.equal(leftHandSide, rightHandSide);
+}
+
+var except = "exception";
+
+var correct_answer_index = 0;
+var correct_answers = [
```
jasongin: Instead of declaring the correct answers like this, why not compare to the results from doing the exact same operations using the JavaScript `instanceof` operator?

test/addons-abi/test_instanceof/test2.js
```diff
@@ -0,0 +1,4 @@
+var addon = require( "./build/Release/test_instanceof" );
+function F() {}
+var x = 1;
+addon.doInstanceOf( x, F );
```
jasongin: I don't understand the purpose of this test file. It doesn't actually assert anything.

test/addons-abi/test_instanceof/test0.js
```diff
@@ -0,0 +1,118 @@
...
+  assertEquals(88, exceptions);
+
+  // Make sure to throw an exception if the function prototype
+  // isn't a proper JavaScript object.
+  function G() { }
+  G.prototype = undefined;
+  assertThrows("(addon.doInstanceOf({}, G))");
+}
+
+TestExceptions();
```
jasongin: This seems redundant with the checks in test1.js that are also verifying exceptional cases.

src/node_jsvmapi.cc
```diff
@@ -815,6 +815,54 @@ napi_value napi_new_instance(napi_env e, napi_value cons,
...
 
+bool napi_instanceof(napi_env e, napi_value obj, napi_value cons) {
+  bool returnValue = false;
+
+  v8::Local<v8::Value> v8Obj = v8impl::V8LocalValueFromJsValue(obj);
+  v8::Local<v8::Value> v8Cons = v8impl::V8LocalValueFromJsValue(cons);
+  v8::Isolate *isolate = v8impl::V8IsolateFromJsEnv(e);
+
+  if (!v8Cons->IsFunction()) {
+	  isolate->ThrowException(
```
jasongin: This could be shortened with `napi_throw_type_error`. Also below.

src/node_jsvmapi.cc
```diff
@@ -815,6 +815,54 @@ napi_value napi_new_instance(napi_env e, napi_value cons,
...
+      v8::String::NewFromUtf8(isolate,
+        "constructor prototype must be an object")));
+
+    // Error handling needs to be done here
+    return false;
+  }
+
+  if (!v8Obj->StrictEquals(v8Cons)) {
+    v8::Local<v8::Value> underunderProto =
+      v8::String::NewFromUtf8(v8impl::V8IsolateFromJsEnv(e), "__proto__");
```
jasongin: So what is the difference between `GetPrototype()` and the `__proto__` property?

src/node_jsvmapi.cc
```diff
@@ -815,6 +815,54 @@ napi_value napi_new_instance(napi_env e, napi_value cons,
...
+  if (!v8Cons->IsObject()) {
+    isolate->ThrowException(v8::Exception::TypeError(
+      v8::String::NewFromUtf8(isolate,
+        "constructor prototype must be an object")));
+
+    // Error handling needs to be done here
+    return false;
+  }
+
+  if (!v8Obj->StrictEquals(v8Cons)) {
```
jasongin, line 842: I don't understand the need for this check, but I assume it is necessary in order to exactly duplicate the semantics of `instanceof`?

gabrielschulhof: Yep, test fails without it :)

src/node_jsvmapi.cc
gabrielschulhof: After testing, it looks like nothing. The only place where it returns something you don't want is when you're using `GetPrototype()` hoping to get the value of the `.prototype` property, which you need [above](https://github.com/nodejs/abi-stable-node/pull/48/files#diff-c148238c1dc2bdc9722b833741c899eeR835).

test/addons-abi/test_instanceof/test0.js
gabrielschulhof: I haven't really checked in detail. I just grabbed all the V8 instanceof-related tests I could get my hands on. I'd keep it, unless we can be absolutely certain that it is a strict subset of the other test.

test/addons-abi/test_instanceof/test2.js
gabrielschulhof: Right, that was just me flexing the addon's muscles :) I'll remove it.

test/addons-abi/test_instanceof/test1.js
gabrielschulhof: Since I copied this test from V8, I decided to change as little as possible. In fact, I'm contemplating loading it dynamically and replacing `x instanceof y ` with `addon.doInstanceOf(x, y)`. I'd prefer this to be part of a well-documented and independently maintained ECMAScript test suite, but alas.

jasongin:
LGTM

Maybe replace all your tabs with spaces though to avoid mixing and be consistent with the Node.js code style.

PR #49, opened 2017-01-16
Try catch 6.2.0 global persistent exception

gabrielschulhof:
Implementation of JSRT exception semantics in V8.

### DIFF
--- a/src/node_jsvmapi.cc
+++ b/src/node_jsvmapi.cc
@@ -86,6 +86,26 @@ namespace v8impl {
     return reinterpret_cast<EscapableHandleScopeWrapper*>(s);
   }
 
+  napi_trycatch JsTryCatchFromBoolPointer(bool* theBoolean) {
+    union U {
+      napi_trycatch t;
+      bool* b;
+      U(bool* _b) : b(_b) { }
+    } u(theBoolean);
+    assert(sizeof(u.t) == sizeof(u.b));
+    return u.t;
+  }
+
+  bool* BoolPointerFromJsTryCatch(napi_trycatch tryCatch) {
+    union U {
+      napi_trycatch t;
+      bool* b;
+      U(napi_trycatch _t) : t(_t) { }
+    } u(tryCatch);
+    assert(sizeof(u.t) == sizeof(u.b));
+    return u.b;
+  }
+
 //=== Conversion between V8 Handles and napi_value ========================
 
   // This is assuming v8::Local<> will always be implemented with a single
@@ -224,6 +244,8 @@ namespace v8impl {
 
 
 }  // end of namespace v8impl
+
+static v8::Persistent<v8::Value> *lastException = 0;
 
 napi_env napi_get_current_env() {
   return v8impl::JsEnvFromV8Isolate(v8::Isolate::GetCurrent());
@@ -580,10 +602,20 @@ napi_value napi_call_function(napi_env e, napi_value scope,
   v8::Local<v8::Function> v8func = v8impl::V8LocalFunctionFromJsValue(func);
   v8::Handle<v8::Object> v8scope =
       v8impl::V8LocalValueFromJsValue(scope)->ToObject();
+  v8::Isolate *isolate = v8impl::V8IsolateFromJsEnv(e);
   for (int i = 0; i < argc; i++) {
     args[i] = v8impl::V8LocalValueFromJsValue(argv[i]);
   }
+  v8::TryCatch tryCatch(isolate);
   v8::Handle<v8::Value> result = v8func->Call(v8scope, argc, args.data());
+  if (tryCatch.HasCaught()) {
+    if (!lastException) {
+      lastException =
+        new v8::Persistent<v8::Value>(isolate, tryCatch.Exception());
+    } else {
+      lastException->Reset(isolate, tryCatch.Exception());
+    }
+  }
   return v8impl::JsValueFromV8LocalValue(result);
 }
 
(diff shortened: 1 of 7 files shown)


gabrielschulhof:
Note: I did not include `napi_trycatch_new()` and `napi_trycatch_delete()` because they would've been noops, just like with ChakraCore. If we decide we need them, I can include them and define them as empty.

src/node_jsvmapi.cc
```diff
@@ -23,6 +23,24 @@
 #include <vector>
 #include <string.h>
 
+static v8::Persistent<v8::Value> *lastException = 0;
+#define NAPI_PROLOGUE \
+  v8::TryCatch napi_internal_tryCatch(v8::Isolate::GetCurrent())
```
jasongin: >v8::Isolate::GetCurrent() [](start = 38, length = 25)

Shouldn't this use the isolate (`napi_env`) that is passed into each function?

src/node_jsvmapi.cc
```diff
@@ -23,6 +23,24 @@
...
+
+#define NAPI_RETURN(returnExpression) \
```
jasongin: >NAPI_RETURN [](start = 8, length = 11)

Could this be implemented as a class with a destructor instead of a macro? Then instead of changing all the `return` statements, the destructor would do this work when the instance goes out of scope.

src/node_jsvmapi.cc
```diff
@@ -186,9 +204,14 @@ namespace v8impl {
                            info.Data().As<v8::External>()->Value());
     napi_func_cb_info cbinfo =
         JsFunctionCallbackInfoFromV8FunctionCallbackInfo(&info);
+    v8::Isolate *isolate = info.GetIsolate();
     cb(
```
jasongin: I think the `lastException` should always be empty at this point before invoking a callback. Would it be helpful to `assert` that?

src/node_jsvmapi.cc
```diff
@@ -186,9 +204,14 @@ namespace v8impl {
                            info.Data().As<v8::External>()->Value());
     napi_func_cb_info cbinfo =
```
jasongin, line 206: Can this code throw a V8 exception? Does it need a `TryCatch`?

src/node_jsvmapi.cc
```diff
@@ -23,6 +23,24 @@
 #include <vector>
 #include <string.h>
 
+static v8::Persistent<v8::Value> *lastException = 0;
```
jasongin: >* [](start = 33, length = 1)

Why a pointer? I think it would be simpler to just make this a static instance.

test/addons-abi/test_exception/test.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+var test_exception = require( "./build/Release/test_exception" );
+var assert = require( "assert" );
+var theError = new Error( "Some error" );
+var throwTheError = function() {
+	throw theError;
+};
```
jasongin, line 8: Add some negative test cases for when an exception is not thrown.

src/node_jsvmapi.cc
gabrielschulhof: Good point!

src/node_jsvmapi.cc
gabrielschulhof: It needs to be constructed with an `Isolate`, which is unavailable at the global scope.

jasongin: Not if it's initially empty. The isolate can be provided later in the call to `Reset()`.

---
In reply to: [96961860](https://github.com/nodejs/abi-stable-node/pull/49#discussion_r96961860) [](ancestors = 96961860)

jasongin: Not if it's initially empty. The isolate can be provided later in the call to `Reset()`.

jasongin: (You could also make it a private static member of the helper class I suggested in the other comment.)

src/node_jsvmapi.cc
gabrielschulhof: That's true. However, I would like to keep `NAPI_PREAMBLE` as a macro, because, when the error handling is implemented, we may need to do more than just declare an instance of that class like, for instance, to bail early if an exception is already pending.

src/node_jsvmapi.cc
gabrielschulhof: I don't believe so, because all it does is cast the data from V8 to NAPI. It doesn't AFAICT do any casting via V8's own casting mechanisms.

test/addons-abi/test_exception/test.js
gabrielschulhof: Done, below.

src/node_jsvmapi.cc
```diff
@@ -836,38 +936,52 @@ napi_value napi_make_callback(napi_env e, napi_value recv,
...
+// Methods to support catching exceptions
+bool napi_is_exception_pending(napi_env) {
+  return !v8impl::TryCatch::lastException().IsEmpty();
+}
+
+napi_value napi_get_and_clear_last_exception(napi_env e) {
+
+  // TODO: Is there a chance that an exception will be thrown in the process of
+  // attempting to retrieve the global static exception?
+  napi_value returnValue = napi_get_undefined_(e);
```
jasongin, line 948: This first call to `napi_get_undefined_()` is unnecessary.

jasongin:
LGTM. I just had the one minor comment about the latest version.

Issue #53, opened 2017-01-20
C++ helper classes


jasongin:
Code that is converted from V8's C++ classes to NAPI's C function  calls is often a little more verbose, e.g. `napi_get_number_from_value(env, v)` vs  `v->NumberValue()`. This could be cleaned up somewhat by adding low/zero-overhead C++ helper classes in NAPI, allowing for more V8 style interaction with objects, and therefore easier porting.

jasongin:
PR at https://github.com/nodejs/abi-stable-node/pull/88

aruneshchandra:
This is landed already 
https://github.com/nodejs/abi-stable-node-wrapper/commit/fe1d9fd5a2fa898f09695faeb40f5b251b162517

PR #58, opened 2017-01-24
Add api calls for supporting node-sqlite3

sampsongao:
Added some api calls for supporting port node-sqlite3. Please review it and tell me if there is redundancy.

The interface keyword is causing failure in build, so changed it to struct.
The initialize list of CallbackWrapperBase has an error when build saying "need a primary expression before >", not sure what cause that so I changed it to the following
```
    CallbackWrapperBase(const T& cbinfo)
        : _cbinfo(cbinfo),
-         _cbdata(cbinfo.Data().As<v8::Object>()) {
+         _cbdata(v8::Local<v8::Object>::Cast(cbinfo.Data())) {
```

### DIFF
--- a/src/node_api_helpers.h
+++ b/src/node_api_helpers.h
@@ -24,6 +24,13 @@
 #define NAPI_MODULE_INIT(name)                                                 \
   void name(napi_env env, napi_value exports, napi_value module)
 
+#define NAPI_GETTER(name)                                                      \
+  void name(napi_env e, napi_propertyname property, napi_callback_info info)
+
+#define NAPI_SETTER(name)                                                      \
+  void name(napi_env e, napi_propertyname property, napi_value value,          \
+            napi_callback_info info)
+
 // This is taken from NAN and is the C++11 version.
 // TODO(ianhall): Support pre-C++11 compilation?
 #define NAPI_DISALLOW_ASSIGN(CLASS) void operator=(const CLASS&) = delete;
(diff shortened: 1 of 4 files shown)


src/node_api_helpers.h
```diff
@@ -24,6 +24,13 @@
 #define NAPI_MODULE_INIT(name)                                                 \
   void name(napi_env env, napi_value exports, napi_value module)
 
+#define NAPI_GETTER(name)                                                      \
+  void name(napi_env e, napi_propertyname property, napi_callback_info info)
+
+#define NAPI_SETTER(name)                                                      \
+  void name(napi_env e, napi_propertyname property, napi_value value,          \
+            napi_callback_info info)
```
jasongin: I'd like to discuss your motivations for this change. These macros are already defined just above, to be the same as `NAPI_METHOD`. Giving them signatures that don't match `napi_callback` is going to cause conflicts with APIs that use `napi_property_descriptor`. When I added the accessors implementation, I considered having accessor callbacks follow this pattern (to more closely follow Nan), but it made the NAPI api more complicated while not actually providing any added functionality.

src/node_jsvmapi.cc
```diff
@@ -864,6 +888,10 @@ napi_value napi_get_true(napi_env e) {
              v8::True(v8impl::V8IsolateFromJsEnv(e)));
 }
 
+napi_value napi_new_value(napi_env env) {
```
jasongin: So.. does this make an _empty_ value? If so, should it be named `napi_new_empty_value`?

src/node_api_helpers.h
```diff
@@ -38,6 +45,289 @@
...
 
 namespace Napi {
+  inline napi_value Value() {
+    napi_env env = napi_get_current_env();
+    return napi_new_value(env);
+  }
+
+
+  // Napi::New helpers
+  inline napi_value New() {
```
jasongin: I can see how all these helpers are useful. [I had been thinking](https://github.com/nodejs/abi-stable-node/issues/53) eventually we might create a C++ wrapper class around `napi_value`, that would have a lot of these sort of helper methods (and more). I guess we can go with this for now, and replace with something richer later.

src/node_api_helpers.h
```diff
@@ -38,6 +45,289 @@
...
+    return napi_string_concat(env, str1, str2);
+  }
+
+  inline napi_value MakeCallback(
+      napi_value target,
+      napi_value func,
+      int argc,
+      napi_value* argv
+  ) {
+    napi_env env = napi_get_current_env();
```
jasongin: All these helpers remind me of something I haven't understood yet... if it is OK to always use the "current" environment (isolate), then why do NAPI functions all have the explicit environment parameter? Nan always just uses the current one, right? In what scenario might a caller of a NAPI function specify an environment that is different from the current one?

src/node_api_helpers.h
```diff
@@ -38,6 +45,289 @@
...
+    return napi_new_value(env);
+  }
+
+
+  // Napi::New helpers
+  inline napi_value New() {
+    return napi_get_null(napi_get_current_env());
+  }
+  inline napi_value New(bool val) {
```
jasongin: Unfortunately most of these helper functions are going to have to change significantly to accommodate the new error-handling pattern. You might want to wait until that change goes in so you can build the helpers on it.

src/node_api_helpers.h
```diff
@@ -38,6 +45,289 @@
...
+
+  inline bool Equals(napi_value lhs, napi_value rhs) {
+    napi_env env = napi_get_current_env();
+    return napi_strict_equals(env, lhs, rhs);
+  }
+
+  inline napi_value* GetArguments(napi_callback_info info) {
+    napi_env env = napi_get_current_env();
+    int len = napi_get_cb_args_length(env, info);
+    napi_value* args = (napi_value*) malloc(len * sizeof(napi_value));
```
jasongin: When does this memory get freed?

It's OK for the helpers to use C++, right? So could this return something like a `std::vector` to make memory-management easier?

src/node_jsvmapi.cc
```diff
@@ -204,7 +204,7 @@ namespace v8impl {
     public:
       CallbackWrapperBase(const T& cbinfo)
         : _cbinfo(cbinfo),
-          _cbdata(cbinfo.Data().As<v8::Object>()) {
+          _cbdata(v8::Local<v8::Object>::Cast(cbinfo.Data())) {
```
jasongin, line 207: Why is a cast necessary? Only our own internal code can invoke this constructor, and it always creates an Object to hold the callback data.

src/node_jsvmapi.h
```diff
@@ -112,13 +112,18 @@ enum napi_valuetype {
   napi_number,
   napi_string,
   napi_symbol,
+  napi_regexp,
+  napi_date,
+  napi_array,
```
jasongin: Array, Date, and RegExp are not fundamental JavaScript types, they are subtypes of Object. I don't think they belong in this enum, or in `napi_get_type_of_value()`. That is why there is a separate `napi_is_array()` function.

src/node_jsvmapi.h
```diff
@@ -151,6 +157,8 @@ NODE_EXTERN napi_value napi_create_type_error(napi_env e, napi_value msg);
 
 // Methods to get the the native napi_value from Primitive type
 NODE_EXTERN napi_valuetype napi_get_type_of_value(napi_env e, napi_value v);
+NODE_EXTERN bool napi_is_int32(napi_env e, napi_value v);
+NODE_EXTERN bool napi_is_uint32(napi_env e, napi_value v);
```
jasongin: While V8 treats certain numbers as int32 or uint32, that is a V8-specific optimization that I'm not sure we should expose in NAPI.

src/node_jsvmapi.h
```diff
@@ -250,6 +260,11 @@ NODE_EXTERN void napi_wrap(napi_env e, napi_value jsObject, void* nativeObj,
                            napi_destruct napi_destructor,
                            napi_weakref* handle);
 NODE_EXTERN void* napi_unwrap(napi_env e, napi_value jsObject);
+
+NODE_EXTERN int napi_get_internal_field_count(napi_env env, napi_value jsObject);
+NODE_EXTERN void* napi_get_internal_field_pointer(napi_env env, napi_value jsObject, int index);
+NODE_EXTERN void napi_set_internal_field_pointer(napi_env env, napi_value jsObject,  int index, void* nativeObj);
```
jasongin: Internal fields are V8-specific, right? I think the uses of internal fields can be handled in other ways with existing NAPI functions, though it may make porting a little more difficult.

src/node_jsvmapi.cc
jasongin: Oh, I see your note in the PR description about this. I think there is a better way to fix this, using the [template disambiguator](http://stackoverflow.com/questions/4077110/template-disambiguator) syntax I haven't used before:
```
cbdata(cbinfo.template Data().As<v8::Object>())
```

src/node_api_helpers.h
gabrielschulhof: IIRC node is working on support for multiple isolates.

src/node_api_helpers.h
mhdawson: In person review with Sampon, he will merge with the above defs.

src/node_jsvmapi.cc
sampsongao: Yes, this is to avoid some cases where module code end up using an napi_value uninitialize and causing a segfault.

src/node_api_helpers.h
mhdawson: I think its important that we always pass the env through our calls.  Therefore we should always already have the env and should be able to avoid the call to napi_get_current_env().  For many of these I think we should look at just doing a replacement from NAN::StringConcat() to napi_string_concat() directly.  It should be pretty easy with script that supports doing this.

mhdawson: It would be useful to anotate or arrange in a way that its easy to tell which helpers are direct clones of functions in NAN.

sampsongao: I will update this with env passed in. These are just making mannual conversion easier.

src/node_api_helpers.h
```diff
@@ -123,6 +413,84 @@ namespace Napi {
...
+        napi_persistent_make_weak(env, persistent(), this);
+    }
+
+    int refs_;
+
+   private:
+    napi_persistent handle_;
+};
+
+
```
mhdawson, line 438: We already have  C api to provide ObjectWrap functionality (as used in existing object wrap examples https://github.com/nodejs/abi-stable-node-addon-examples/blob/master/6_object_wrap/nan/myobject.h ) .  I think we should re-use that instead of re-creating. If there is something missing then we should add to that existing API.

src/node_api_helpers.h
```diff
@@ -398,9 +766,9 @@ namespace Napi {
     napi_async_set_data(req, static_cast<void*>(worker));
     napi_async_set_execute(req, &AsyncWorker::CallExecute);
     napi_async_set_complete(req, &AsyncWorker::CallWorkComplete);
-    napi_async_set_destroy(req, &AsyncWorker::CallDestroy);
     napi_async_queue_worker(req);
   }
+
```
mhdawson: Do we need this extra line

src/node_jsvmapi.cc
mhdawson: I agree, making it explicit that it returns an empty value would be good.

src/node_jsvmapi.cc
```diff
@@ -1071,6 +1108,21 @@ void* napi_unwrap(napi_env e, napi_value jsObject) {
...
+void* napi_get_internal_field_pointer(napi_env env, napi_value jsObject, int index) {
+  v8::Local<v8::Object> obj = v8impl::V8LocalValueFromJsValue(jsObject).As<v8::Object>();
+  return obj->GetAlignedPointerFromInternalField(index);
+}
+
+void napi_set_internal_field_pointer(napi_env env, napi_value jsObject, int index, void* nativeObj) {
+  v8::Local<v8::Object> obj = v8impl::V8LocalValueFromJsValue(jsObject).As<v8::Object>();
+  obj->SetAlignedPointerInInternalField(index, nativeObj);
+}
+
```
mhdawson: @boingoing can you confirm that chakra has the equivalent of internal fields, I'm a bit worried this might be v8 specific.

src/node_jsvmapi.h
mhdawson: agreed.

src/node_jsvmapi.h
mhdawson: I made a comment above raising a similar concern.  We should see how we can avoid use of the internal fields as our first choice.

src/node_jsvmapi.cc
sampsongao: The .As<v8::Object> is doing the cast in its implementation in v8. So I would prefer to not to use template keyword here.

src/node_jsvmapi.h
kkoopa: Might V8's internal fields correspond to 'Object internal slots'?
http://www.ecma-international.org/ecma-262/6.0/#sec-object-internal-methods-and-internal-slots

jasongin:
This PR is very out of date now after the error-handling changes and the C++ wrapper PR that does some of the same things as this but is far more complete.

@sampsongao Do you want to just close it, or update it to just the things you need after the other changes?

src/node_api_helpers.h
kkoopa: I'll restate this here for future reference: It is not ok to always get the current Isolate. The reason NAN does it is because of legacy support: Before Node 0.12, there was no support for multiple `Isolate`s. Since NAN tries to guarantee identical behavior on all supported platforms, it does not support multiple Isolates.

At some point the V8 team have stated that they would like to completely remove `Isolate::GetCurrent()`. I would therefore advise completely removing the functions for getting the current `Isolate`, else you will be stuck with reimplementing and maintaining it.

jasongin: I opened a new issue about removing access to the current isolate: https://github.com/nodejs/abi-stable-node/issues/96

@kkoopa Also note this PR is obsolete - see [my comment about that](https://github.com/nodejs/abi-stable-node/pull/58#issuecomment-278740450).

kkoopa: @jasongin Thanks for the heads up. Fortunately I did not yet review this thoroughly, I merely looked through the comments.

aruneshchandra:
Not needed at this time.

Issue #37, opened 2017-01-06
Design for NAPI error handling 


aruneshchandra:
Publish and review design document for NAPI error handling

aruneshchandra:
The design needs to also address #27 Correct abstraction for try/catch

boingoing:
Node.js API doesn't yet handle any of the errors which may be encountered within the API itself. Some of the errors can cause AV while others can silently fail and the consumer of the API would not know if the operation was successful. We need to consider what degree of handling we want to implement. Should the API be paranoid about incoming pointers or skip null checks in favor of informing the caller via AV at runtime? How is an error reported to the user of the API and how detailed is that error? This document attempts to outline the issues we need to consider and provides a rough first-pass guide to the design.

To answer the question of how to convey the presence of an error we have a couple of design options:
- Flat error codes like enum or constant integer values as return from API calls
- Optional out-parameter added to each API for error information passed via a structure similar to nserror in iOS development in Objective-C
- Throw exceptions from inside API calls in error cases

Throwing exceptions is a non-starter since we want this API to be flat C. Adding C++ exceptions there would make the API unstable. It's possible for a C++ wrapper of Node.js API to consume one of the other types of error information and throw exceptions. This would not be internal to NAPI, though, which puts it outside the scope of this document.

Optional-out parameter for extra error information seems plausible. More information can flow upstream to provide better error information - perhaps via a future C++ wrapper around NAPI which supports throwing exceptions up to native module code. Concern that the structure of the container for extra information and the added parameter for every API call will make breakages of NAPI more common. 

Similar to the structure of a container for extra error information, concern is also raised over a fixed set of error codes - how do we extend this without breaking the API? Are we safe if the set of error codes is strictly additive? Can we retire error codes?

For the purpose of this document, we will present a design where errors are communicated via simple, integer return values with more error details available via an API call. This design is based on the Node 6 branch of abi-stable-node.

The work in progress for this design is in this fork of abi-stable-node - https://github.com/boingoing/abi-stable-node/tree/error-handling-prototype-6.2.0.

```c++
enum napi_errorvalue {
  napi_ok,
  napi_invalid_arg,
...
  napi_object_expected,
  napi_generic_failure
};

struct napi_extended_error_info {
  const char* error_message;
  void* engine_reserved;
  uint32_t engine_error_code;
  napi_errorvalue error_code;
};

const napi_extended_error_info* napi_get_last_error_info();

// napi_has_element currently (no handling of errors):
[...]

PR #59, opened 2017-01-25
Error handling prototype 6.2.0

boingoing:
This is the initial commit of the error handling prototype into the 6.2 NAPI branch. It instruments most of the APIs to handle and indicate failure states. 

Please see https://github.com/nodejs/abi-stable-node/issues/37 for a discussion of the design.

### DIFF
--- a/src/node_jsvmapi.cc
+++ b/src/node_jsvmapi.cc
@@ -478,8 +478,9 @@ napi_extended_error_info static_last_error;
 // Warning: Keep in-sync with napi_errorvalue enum
 const char* error_messages[] = {
   NULL,
-  "Invalid pointer passed to NAPI",
+  "Invalid pointer passed as argument",
   "An object was expected",
+  "A string was expected",
   "Unknown failure"
 };
 
@@ -622,20 +623,54 @@ void napi_set_return_value(napi_env e,
   info->SetReturnValue(v);
 }
 
-napi_propertyname napi_property_name(napi_env e, const char* utf8name) {
+napi_errorvalue napi_property_name(napi_env e, 
+                                   const char* utf8name, 
+                                   napi_propertyname* result) {
   NAPI_PREAMBLE(e);
-  v8::Local<v8::String> namestring =
-      v8::String::NewFromUtf8(v8impl::V8IsolateFromJsEnv(e), utf8name,
-          v8::NewStringType::kInternalized).ToLocalChecked();
-  return reinterpret_cast<napi_propertyname>(
-      v8impl::JsValueFromV8LocalValue(namestring));
+
+  if (!result) {
+    return napi_set_last_error_code(napi_invalid_arg);
+  }
+
+  auto maybe = v8::String::NewFromUtf8(v8impl::V8IsolateFromJsEnv(e), utf8name, 
+    v8::NewStringType::kInternalized);
+
+  // Is this an OOM?
+  if (maybe.IsEmpty())
+  {
+    return napi_set_last_error_code(napi_generic_failure);
+  }
+
+  *result = reinterpret_cast<napi_propertyname>(
+    v8impl::JsValueFromV8LocalValue(maybe.ToLocalChecked()));
+
+  return napi_ok;
 }
 
-napi_value napi_get_propertynames(napi_env e, napi_value o) {
+napi_errorvalue napi_get_propertynames(napi_env e, napi_value o, napi_value* result) {
   NAPI_PREAMBLE(e);
-  v8::Local<v8::Object> obj = v8impl::V8LocalValueFromJsValue(o)->ToObject();
-  v8::Local<v8::Array> array = obj->GetPropertyNames();
-  return v8impl::JsValueFromV8LocalValue(array);
+
+  if (!result) {
+    return napi_set_last_error_code(napi_invalid_arg);
+  }
+
+  v8::Isolate *isolate = v8impl::V8IsolateFromJsEnv(e);
+  v8::Local<v8::Context> context = isolate->GetCurrentContext();
+  v8::MaybeLocal<v8::Object> obj_maybe = v8impl::V8LocalValueFromJsValue(o)->ToObject(context);
+
+  if (obj_maybe.IsEmpty()) {
+    return napi_set_last_error_code(napi_object_expected);
+  }
+
+  auto maybe_propertynames = obj_maybe.ToLocalChecked()->GetPropertyNames(context);
+
+  if (maybe_propertynames.IsEmpty())
+  {
+    return napi_set_last_error_code(napi_generic_failure);
+  }
+
+  *result = v8impl::JsValueFromV8LocalValue(maybe_propertynames.ToLocalChecked());
+  return napi_ok;
 }
 
 napi_errorvalue napi_set_property(napi_env e, 
@@ -643,16 +678,21 @@ napi_errorvalue napi_set_property(napi_env e,
                                   napi_propertyname k, 
                                   napi_value v) {
   NAPI_PREAMBLE(e);
-  v8::Local<v8::Object> obj = v8impl::V8LocalValueFromJsValue(o)->ToObject();
-  v8::Local<v8::Value> key = v8impl::V8LocalValueFromJsPropertyName(k);
-  v8::Local<v8::Value> val = v8impl::V8LocalValueFromJsValue(v);
 
   v8::Isolate *isolate = v8impl::V8IsolateFromJsEnv(e);
   v8::Local<v8::Context> context = isolate->GetCurrentContext();
+  v8::MaybeLocal<v8::Object> obj_maybe = v8impl::V8LocalValueFromJsValue(o)->ToObject(context);
 
-  v8::Maybe<bool> maybe = obj->Set(context, key, val);
+  if (obj_maybe.IsEmpty()) {
+    return napi_set_last_error_code(napi_object_expected);
+  }
 
-  if (!maybe.FromMaybe(false)) {
+  v8::Local<v8::Value> key = v8impl::V8LocalValueFromJsPropertyName(k);
+  v8::Local<v8::Value> val = v8impl::V8LocalValueFromJsValue(v);
+
+  v8::Maybe<bool> set_maybe = obj_maybe.ToLocalChecked()->Set(context, key, val);
+
+  if (!set_maybe.FromMaybe(false)) {
     return napi_set_last_error_code(napi_generic_failure);
   }
 
@@ -761,26 +801,57 @@ napi_errorvalue napi_has_element(napi_env e, napi_value o, uint32_t i, bool* has
   return napi_ok;
 }
 
-napi_value napi_get_element(napi_env e, napi_value o, uint32_t i) {
+napi_errorvalue napi_get_element(napi_env e, 
+                                 napi_value o, 
+                                 uint32_t i, 
+                                 napi_value* result) {
   NAPI_PREAMBLE(e);
 
-  v8::Local<v8::Object> obj = v8impl::V8LocalValueFromJsValue(o)->ToObject();
-  v8::Local<v8::Value> val = obj->Get(i);
-  // This implementation is missing a lot of details, notably error
-  // handling on invalid inputs and regarding what happens in the
-  // Set operation (error thrown, key is invalid, the bool return
-  // value of Set)
-  return v8impl::JsValueFromV8LocalValue(val);
-}
+  if (!result) {
+    return napi_set_last_error_code(napi_invalid_arg);
+  }
 
-void napi_define_property(napi_env e, napi_value o,
-    napi_property_descriptor* p) {
   v8::Isolate *isolate = v8impl::V8IsolateFromJsEnv(e);
   v8::Local<v8::Context> context = isolate->GetCurrentContext();
-  v8::Local<v8::Object> obj = v8impl::V8LocalValueFromJsValue(o)->ToObject();
-  v8::Local<v8::Name> name =
-    v8::String::NewFromUtf8(isolate, p->utf8name,
-      v8::NewStringType::kInternalized).ToLocalChecked();
+  auto obj_maybe = v8impl::V8LocalValueFromJsValue(o)->ToObject(context);
+
+  if (obj_maybe.IsEmpty()) {
+    return napi_set_last_error_code(napi_object_expected);
+  }
+
+  auto get_maybe = obj_maybe.ToLocalChecked()->Get(context, i);
+
+  if (get_maybe.IsEmpty()) {
+    return napi_set_last_error_code(napi_generic_failure);
+  }
+
+  *result = v8impl::JsValueFromV8LocalValue(get_maybe.ToLocalChecked());
+  return napi_ok;
+}
+
+napi_errorvalue napi_define_property(napi_env e, napi_value o,
+    napi_property_descriptor* p) {
+  NAPI_PREAMBLE(e);
+
+  v8::Isolate *isolate = v8impl::V8IsolateFromJsEnv(e);
+  v8::Local<v8::Context> context = isolate->GetCurrentContext();
+  auto obj_maybe = v8impl::V8LocalValueFromJsValue(o)->ToObject(context);
+
+  if (obj_maybe.IsEmpty()) {
+    return napi_set_last_error_code(napi_object_expected);
+  }
+
+  auto obj = obj_maybe.ToLocalChecked();
+
+  auto name_maybe = v8::String::NewFromUtf8(isolate, p->utf8name,
+    v8::NewStringType::kInternalized);
+
+  if (name_maybe.IsEmpty())
+  {
+    return napi_set_last_error_code(napi_generic_failure);
+  }
+
+  v8::Local<v8::Name> name = name_maybe.ToLocalChecked();
 
   v8::PropertyAttribute attributes =
     static_cast<v8::PropertyAttribute>(p->attributes);
@@ -792,17 +863,24 @@ void napi_define_property(napi_env e, napi_value o,
     v8::Local<v8::FunctionTemplate> t = v8::FunctionTemplate::New(
       isolate, v8impl::FunctionCallbackWrapper::Invoke, cbdata);
 
-    obj->DefineOwnProperty(
+    auto define_maybe = obj->DefineOwnProperty(
       context,
       name,
       t->GetFunction(),
       attributes);
+
+    // IsNothing seems like a serious failure, 
+    // should we return a different error code if the define failed?
+    if (define_maybe.IsNothing() || !define_maybe.FromMaybe(false))
+    {
+      return napi_set_last_error_code(napi_generic_failure);
+    }
   }
   else if (p->getter || p->setter) {
     v8::Local<v8::Object> cbdata = v8impl::CreateAccessorCallbackData(
       e, p->getter, p->setter, p->data);
 
-    obj->SetAccessor(
+    auto set_maybe = obj->SetAccessor(
       context,
       name,
       v8impl::GetterCallbackWrapper::Invoke,
@@ -810,15 +888,32 @@ void napi_define_property(napi_env e, napi_value o,
       cbdata,
       v8::AccessControl::DEFAULT,
       attributes);
+
+    // IsNothing seems like a serious failure, 
+    // should we return a different error code if the set failed?
+    if (set_maybe.IsNothing() || !set_maybe.FromMaybe(false))
+    {
+      return napi_set_last_error_code(napi_generic_failure);
+    }
   }
   else {
     v8::Local<v8::Value> value = v8impl::V8LocalValueFromJsValue(p->value);
-    obj->DefineOwnProperty(
+
+    auto define_maybe = obj->DefineOwnProperty(
       context,
       name,
       value,
       attributes);
+
+    // IsNothing seems like a serious failure, 
+    // should we return a different error code if the define failed?
+    if (define_maybe.IsNothing() || !define_maybe.FromMaybe(false))
+    {
+      return napi_set_last_error_code(napi_generic_failure);
+    }
   }
+
+  return napi_ok;
 }
 
 bool napi_is_array(napi_env e, napi_value v) {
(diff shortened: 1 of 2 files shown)


src/node_jsvmapi.h
```diff
@@ -116,80 +116,99 @@ enum napi_valuetype {
   napi_function,
 };
 
-// Environment
-NODE_EXTERN napi_env napi_get_current_env();
+enum napi_errorvalue {
```
jasongin: I'd suggest `napi_status` as a simpler name. (It's not usually an error anyway.)

jasongin: In the design discussion we had talked about a flag of some sort in the return code that would additionally indicate whether an exception is pending. I can't remember what the consensus was on that.

src/node_jsvmapi.cc
```diff
@@ -465,22 +465,58 @@ namespace v8impl {
...
+  NULL,
+  "Invalid pointer passed as argument",
+  "An object was expected",
+  "A string was expected",
+  "Unknown failure"
+};
+
+const napi_extended_error_info* napi_get_last_error_info() {
+  // Wait until someone requests the last error information to fetch the error message string
+  static_last_error.error_message = error_messages[static_last_error.error_code];
```
jasongin, line 544: Should probably check that the error code is within range, and fallback to null or a default message if it's not. (Yeah as long as it's one of the enum values it *should* be in range, but it would be easy to add an enum value without adding a string, or pass in some other value to the function.)

src/node_jsvmapi.cc
```diff
@@ -465,22 +465,58 @@ namespace v8impl {
...
+
+  return &static_last_error;
 }
 
-napi_value napi_create_function(
+napi_errorvalue napi_set_last_error_code(napi_errorvalue error_code) {
+  static_last_error.error_code = error_code;
```
jasongin, line 552: I'm concerned that by only setting `error_code`, this makes it easy for other fields (the engine-specific info) to "leak" across different error instances since it never gets reset.

src/node_jsvmapi.h
```diff
@@ -116,80 +116,99 @@ enum napi_valuetype {
...
+  napi_ok,
+  napi_invalid_arg,
+  napi_object_expected,
+  napi_string_expected,
+  napi_generic_failure
+};
 
+struct napi_extended_error_info {
+  const char* error_message;
+  void* engine_reserved;
```
jasongin, line 131: I still wonder if we need a way to free whatever this might point to. (Or should it be a napi_value?)

src/node_jsvmapi.cc
```diff
@@ -465,22 +465,58 @@ namespace v8impl {
...
 
-// We define a macro here because, with full error handling, we may want to
-// perform additional actions at the top of our public APIs, such as bail,
-// returning an error code at the very top of each public API in the case that
-// there's already an exception pending in v8impl::TryCatch::lastException
+   // We define a macro here because, with full error handling, we may want to
+   // perform additional actions at the top of our public APIs, such as bail,
+   // returning an error code at the very top of each public API in the case that
+   // there's already an exception pending in v8impl::TryCatch::lastException
 #define NAPI_PREAMBLE(e) \
```
jasongin: Should this macro now check if `e` is null and return an error if it is?

src/node_jsvmapi.cc
```diff
@@ -465,22 +465,58 @@ namespace v8impl {
...
     napi_env e,
     napi_callback cb,
-    void* data) {
+    void* data,
+    napi_value* result) {
   NAPI_PREAMBLE(e);
+
+  if (!result) {
+    return napi_set_last_error_code(napi_invalid_arg);
+  }
```
jasongin: Should there be a macro for this check, since it's repeated for nearly every API?

src/node_jsvmapi.cc
```diff
@@ -735,133 +936,257 @@ bool napi_strict_equals(napi_env e, napi_value lhs, napi_value rhs) {
...
+  *result = v8impl::JsValueFromV8LocalValue(
       v8::Exception::TypeError(
           v8impl::V8LocalValueFromJsValue(msg).As<v8::String>()));
+
+  return napi_ok;
 }
 
-napi_valuetype napi_get_type_of_value(napi_env e, napi_value vv) {
-  v8::Local<v8::Value> v = v8impl::V8LocalValueFromJsValue(vv);
+napi_errorvalue napi_get_type_of_value(napi_env e, napi_value vv, napi_valuetype* valuetype) {
```
jasongin: Change `valuetype` to `result` for consistency?

src/node_jsvmapi.cc
```diff
@@ -585,96 +623,235 @@ void napi_set_return_value(napi_env e,
...
 
-  // This implementation is missing a lot of details, notably error
-  // handling on invalid inputs and regarding what happens in the
-  // Set operation (error thrown, key is invalid, the bool return
-  // value of Set)
+  return napi_ok;
 }
 
-bool napi_has_property(napi_env e, napi_value o, napi_propertyname k) {
+napi_errorvalue napi_has_property(napi_env e, napi_value o, napi_propertyname k, bool* has) {
```
jasongin: Change `has` to `result` for consistency?

src/node_jsvmapi.cc
```diff
@@ -585,96 +623,235 @@ void napi_set_return_value(napi_env e,
...
+
+  auto obj = obj_maybe.ToLocalChecked();
+
+  auto name_maybe = v8::String::NewFromUtf8(isolate, p->utf8name,
+    v8::NewStringType::kInternalized);
+
+  if (name_maybe.IsEmpty())
+  {
+    return napi_set_last_error_code(napi_generic_failure);
+  }
```
jasongin: I wonder if we should define a `CHECK_MAYBE_EMPTY` macro or something like that to streamline this repeated pattern.

src/node_jsvmapi.cc
gabrielschulhof: That was my intention for when error handling is in place.

jasongin: Should it also *clear* the last error code? For reference, Windows' `GetLastError()` always returns 0 after a return from a successful API call (even after there was a previous error).

src/node_jsvmapi.cc
```diff
@@ -585,96 +623,235 @@ void napi_set_return_value(napi_env e,
...
-      v8impl::JsValueFromV8LocalValue(namestring));
+
+  if (!result) {
+    return napi_set_last_error_code(napi_invalid_arg);
+  }
+
+  auto maybe = v8::String::NewFromUtf8(v8impl::V8IsolateFromJsEnv(e), utf8name, 
+    v8::NewStringType::kInternalized);
+
+  // Is this an OOM?
```
jasongin: I'm not sure, but this makes me wonder about other V8 new functions like `v8::Array::New()` which don't return `Maybe`s. Do they set a pending exception if they fail?

src/node_jsvmapi.cc
gabrielschulhof: glib has macros similar to this one:
```C
#define RETURN_VALUE_IF_FAIL(condition, value)  \
  do {                                          \
    if (!(condition)) {                         \
      return napi_set_last_error_code((value)); \
    }                                           \
  } while(0)
```

src/node_jsvmapi.cc
gabrielschulhof: Yes, it should. Additionally, if there's a pending exception, it should bail early. So, with the macro I proposed [below](https://github.com/nodejs/abi-stable-node/pull/59#discussion_r97736487), this would look like
```C++
#define NAPI_PREAMBLE(e)                                            \
do {                                                                \
  RETURN_VALUE_IF_FAIL(v8impl::TryCatch::lastException().IsEmpty(), \
    napi_pending_exception);                                        \
  napi_clear_last_error_code();                                     \
  v8impl::TryCatch tryCatch(v8impl::V8IsolateFromJsEnv((e)));       \
} while(0)
```
So, if you call an API, that throws an exception, and you call another API, you lose whatever the original error code was, and the new error code becomes `napi_exception_pending`, and that's what you're gonna get from every subsequent API call.

BTW, I'm also proposing a new `napi_errorvalue` here :)

src/node_jsvmapi.h
boingoing: I can get behind that. napi_errorvalue is just a placeholder for a better name, anyway.

boingoing: I think everyone liked the idea of using some bits to indicate status like this but we didn't settle on if we should actually do it or if it was practical or not. Hence, I haven't done it here. I suppose we could use a macro like NAPI_EPILOGUE to always check the pending exception support?

gabrielschulhof:
Overall, I think this is OK, but I was under the impression that we wouldn't change the methods' return values, but rather we would allow for them to be interpreted in light of a last error/pending exception, which becomes accessible by the new API which we add. Thus, if the last error is set and/or there is a pending exception, then the return value is undefined (not in the Javascript sense, but in the sense of undefined behaviour). I guess we could provide sane return values even if we document that they are undefined in the above-mentioned sense if, in the case of a `napi_value` or a pointer-type return value, the return value would be a `nullptr`, and in the case of an integer-like return value this would be `0` - either way, the initialization value for a variable holding the value we intend to return upon successful completion.

src/node_jsvmapi.cc
boingoing: That's a good idea. I'll add a napi_status_last enum value to the end of the enum to use for the valid range. That wouldn't guard against someone adding a new enum value in the middle without adding a new message in the string array. Maybe we can add a compile assert to the napi_extended_error_info ctor which verifies the count of strings or something like this?

src/node_jsvmapi.cc
gabrielschulhof: `Local` has an [empty-check](https://v8docs.nodesource.com/node-7.4/de/deb/classv8_1_1_local.html#a821ace5666d33802aff59f69f72901dd) as well, so **maybe** ;) we should use that.

src/node_jsvmapi.cc
gabrielschulhof: ... though I guess if we use an extra bit, we don't need the extra enum value, but rather we would clear the lower bits and retain the high bit.

src/node_jsvmapi.cc
boingoing: Yes, that's true. This PR doesn't include napi_clear_extended_error_info or any management of the engine-specific fields. I suppose this function could take the other values as well with defaults set to nullptr or something.

src/node_jsvmapi.h
boingoing: I think void* is right for now until we find a real way for an engine to use it. And, yeah, we definitely at least need napi_clear_extended_error_info.

src/node_jsvmapi.cc
boingoing: I like this macro. Also not sure about using an enum or bit flag for the pending exception, though. Should we always clear the error code here or clear it when we return napi_ok from the API (convert all returns from API to first set the error code before returning it) ?

src/node_jsvmapi.cc
boingoing: Yes, there are a couple of these patterns which should definitely be replaced by a macro.

src/node_jsvmapi.cc
boingoing: I think so, at least for the v8 ports of NAPI.

src/node_jsvmapi.cc
boingoing: It's not clear to me that these new operations are able to call back into script, though I guess they would return an empty Local in OOM cases. I wonder if this changes with ES6 new semantics since you might be able to override default Array creation (for example) via @@species.

src/node_jsvmapi.cc
gabrielschulhof: I think we should clear the error code at the top. Every API call should start with tabula rasa and should short-circuit if there's an exception pending. If everything is fine, it should be sufficient to just `return napi_ok;` without having to do an epilogue. For everything else, there's `RETURN_VALUE_IF_FAIL()`.

... and if we go with our existing signatures and not with `napi_errorvalue` return values, we can simply intersperse `RETURN_VALUE_IF_FAIL(something, nullptr|false|0)` throughout the code.

src/node_jsvmapi.cc
jasongin: Yeah, something like `static_assert(sizeof(error_messages)/sizeof(*error_messages) == napi_status_last)` would be a good check.

jasongin:
> I was under the impression that we wouldn't change the methods' return values, but rather we would allow for them to be interpreted in light of a last error/pending exception, which becomes accessible by the new API which we add. Thus, if the last error is set and/or there is a pending exception, then the return value is undefined (not in the Javascript sense, but in the sense of undefined behaviour).

I think we do need to have all the APIs return a status code. Otherwise, a conscientious programmer will need to call `napi_get_last_status()` *after every API call*, which would certainly impact performance because it can't be inlined. Instead they should always check the returned status code. Probably we should define macros in the NAPI public headers to help make that checking less verbose.

src/node_jsvmapi.h
jasongin: Thinking about this more, I'm not sure a bit flag would be necessary, or worth the added complexity. If there is a status code like `napi_pending_exception`, then we can return that code when there is a pending exception and no other more-specific error conditions were detected.

If the return code is `napi_ok`, then it should be safe for the caller to assume there is no pending exception, assuming there is code in the NAPI implementation to check for a pending exception before returning.

If the return code is some error other than `napi_pending_exception`, then I think the caller cannot make any assumption about whether there is or is not a pending exception; they'll just have to call `napi_is_exception_pending()`. Is that OK, or should we try to make some sort of guarantee?

Checking for a pending exception *from within NAPI code* should be very fast: both the retrieval of V8's static `lastException` persistent value and the `IsEmpty()` check on it should be inlined. (But that won't be the case when calling across the NAPI boundary.)

mhdawson: The bit would allow you to avoid the call to napi_is_exception_pending in the case were it is useful to return a different error code.  

I don't think its all that hard to have some macros to mask out the bit/check if there was an exception.

src/node_jsvmapi.cc
mhdawson: @gabrielschulhof as an aside did you find documentation for TryCatch.lastException()  or just look at the code.  Was looking for it to be able to complete my understanding of the exception code but have not found doc through google yet.

mhdawson:
I'm in agreement that we do want all APIs return a status code.  There should be no requirement to call napi_get_last_status unless you want more detailed info.

src/node_jsvmapi.h
jasongin: @mhdawson In my (admittedly limited) experience reviewing native module code, it's fairly unusual for the native module to care whether or not there is a pending exception. Most of the time, they can just let it remain pending so that it gets thrown whenever the native call returns to JS. Therefore, I think it might not be worth the added complexity to have status codes that are a combination of two things instead of a simple enum.

src/node_jsvmapi.cc
```diff
@@ -465,22 +465,109 @@ namespace v8impl {
 
 }  // end of namespace v8impl
 
-// We define a macro here because, with full error handling, we may want to
-// perform additional actions at the top of our public APIs, such as bail,
-// returning an error code at the very top of each public API in the case that
-// there's already an exception pending in v8impl::TryCatch::lastException
-#define NAPI_PREAMBLE(e) \
-  v8impl::TryCatch tryCatch(v8impl::V8IsolateFromJsEnv((e)))
+#define RETURN_STATUS_IF_FAIL(condition, status)                        \
```
jasongin: I'd name this `RETURN_STATUS_IF_FALSE` or `..._IF_NOT`, since really it checks if the condition is false, not if it "failed".

src/node_jsvmapi.cc
```diff
@@ -465,22 +465,109 @@ namespace v8impl {
...
+
+#define CHECK_MAYBE_NOTHING(maybe, status)                              \
+  RETURN_STATUS_IF_FAIL(!((maybe).IsNothing()), (status))
+
+#define NAPI_PREAMBLE(e)                                                \
+  do {                                                                  \
+    RETURN_STATUS_IF_FAIL(v8impl::TryCatch::lastException().IsEmpty(),  \
+      napi_pending_exception);                                          \
+    napi_clear_last_error_code();                                       \
+    v8impl::TryCatch tryCatch(v8impl::V8IsolateFromJsEnv((e)));         \
```
jasongin, line 489: `CHECK_ARG(e)` before this? It looks like `v8::TryCatch` constructor doesn't validate the isolate, so we'll need to if we want to avoid an AV.

src/node_jsvmapi.cc
```diff
@@ -465,22 +465,109 @@ namespace v8impl {
...
 
-napi_env napi_get_current_env() {
-  return v8impl::JsEnvFromV8Isolate(v8::Isolate::GetCurrent());
+  // Wait until someone requests the last error information to fetch the error message string
+  static_last_error.error_message = error_messages[static_last_error.error_code];
+
+  return &static_last_error;
+}
+
+void napi_clear_extended_error_info() {
```
jasongin: I think it would be simpler to combine `napi_clear_extended_error_info()` and `napi_clear_last_error_code()` in a single `napi_clear_last_error()` that clears everything. I can't think of a scenario where it would be necessary to clear only one part or the other.

src/node_jsvmapi.cc
```diff
@@ -465,22 +465,109 @@ namespace v8impl {
...
+  // TODO: Should this be a callback?
+  static_last_error.engine_error_code = 0;
+  static_last_error.engine_reserved = NULL;
+}
+
+napi_status napi_set_last_error_code(napi_status error_code,
```
jasongin: Maybe shorten this to `napi_set_last_error` now since it can set more than just the code.

src/node_jsvmapi.h
```diff
@@ -288,13 +321,17 @@ NODE_EXTERN bool napi_is_exception_pending(napi_env e);
 NODE_EXTERN napi_value napi_get_and_clear_last_exception(napi_env e);
```
jasongin: Above methods (for handle scopes and exceptions) still need to be updated to return `napi_status`. You were probably aware, but I'm flagging this since it's the only remaining change needed in the public API.

src/node_jsvmapi.cc
```diff
@@ -465,22 +465,109 @@ namespace v8impl {
...
+
+#define CHECK_NEW_FROM_UTF8(result, str)                                \
+    CHECK_NEW_FROM_UTF8_LEN((result), (str), -1)
+
+   // Static last error returned from napi_get_last_error_info
+napi_extended_error_info static_last_error;
+
+// Warning: Keep in-sync with napi_status enum
+const char* error_messages[] = {
+  NULL,
```
jasongin: To be consistent, use `nullptr` instead of `NULL`. Also in a few other places.

src/node_jsvmapi.cc
```diff
@@ -465,22 +465,109 @@ namespace v8impl {
...
+
+#define NAPI_PREAMBLE(e)                                                \
+  do {                                                                  \
+    RETURN_STATUS_IF_FAIL(v8impl::TryCatch::lastException().IsEmpty(),  \
+      napi_pending_exception);                                          \
+    napi_clear_last_error_code();                                       \
+    v8impl::TryCatch tryCatch(v8impl::V8IsolateFromJsEnv((e)));         \
+  } while(0)
+
+#define CHECK_TO_OBJECT(result, o)                                      \
```
jasongin: We could probably use a `ToString()` flavor of this also. Or perhaps use preprocessor token pasting to make macro that works with any type:
```
#define CHECK_TO_X(type, result, o)  \
    ...
    auto maybe = v8impl::V8LocalValueFromJsValue((o))->To##type(context);
    ...

#define CHECK_TO_STRING(result, o)  CHECK_TO_X(String, result, o)
```

jasongin:
Great progress! This is getting close. I added more comments on the latest revision, but they're mostly about very minor details. The biggest piece of remaining work is the updates to the handle scope and exception APIs.

src/node_jsvmapi.cc
boingoing: I made the change to nullptr in those places, but I believe there was some resistance earlier in the project due to C linkage.

src/node_jsvmapi.cc
boingoing: This makes sense and I did refactor it but we only use ToString in one place, so I do wonder at how useful that generalization is. Still a good idea :)

src/node_jsvmapi.h
boingoing: Yes, I knew about these and I skipped them so we could talk about whether or not we should instrument them to return napi_status. Went ahead and did instrument them after our meeting Friday. Sounds like the right thing to do.

src/node_jsvmapi.cc
jasongin: I think it shouldn't be an issue as long as it's in the .cc file. I suppose we do want to keep the .h file C-compatible though.

jasongin:
LGTM!

jasongin:
The test cases need to be updated for the API changes; I told Taylor I'd do that separately. I should have time tomorrow (Monday).

mhdawson:
@boingoing  think we'd agreed you were going to integrate a check when the return codes were generated so that napi_pending_exception would be returned if there was an exception pending and no other error code needed to be returned.   The only place I can find it being set is:
```
#define NAPI_PREAMBLE(e)                                                \
  do {                                                                  \
    CHECK_ARG(e);                                                       \
    RETURN_STATUS_IF_FALSE(v8impl::TryCatch::lastException().IsEmpty(), \
      napi_pending_exception);                                          \
    napi_clear_last_error();                                            \
    v8impl::TryCatch tryCatch(v8impl::V8IsolateFromJsEnv((e)));         \
  } while(0)
```

which will catch the case when we make a napi call when an exception is pending. But it would not set the return code when we are returning from a napi call and an exception is pending.  

Am I missing something ?

mhdawson:
@boingoing, @aruneshchandra  I see you left this in:
```
+ *  - The V8 implementation of the API is roughly hacked together with only basic
- *    attention paid to error handling or fault tolerance.
+ *    error handling or fault tolerance._
```
Does that mean you think there is still significant work to properly do error handling ? If so I'm wondering if we really need to do that before performance analysis.

boingoing:
@mhdawson 

> Does that mean you think there is still significant work to properly do error handling ? If so I'm wondering if we really need to do that before performance analysis.

I didn't feel comfortable just deleting all these lines of the comment since we are still working on a prototype, right? I changed the comment from "nothing has been done" to "basic attention has been paid" which seems closer. I don't think there is significant work left to do proper error handling, but there will be some tail of issues we need to finish.

boingoing:
@mhdawson 

> think we'd agreed you were going to integrate a check when the return codes were generated so that napi_pending_exception would be returned if there was an exception pending and no other error code needed to be returned.

Yes, I didn't add that into the initial commit over the weekend but I am implementing it today. I'll send a PR with that change.

boingoing:
@mhdawson 

Sent PR https://github.com/nodejs/abi-stable-node/pull/73 for this issue.

mhdawson:
@boingoing I think we'll have tail issues across the board so I'm thinking that unless we think there is significant/major work left on the error handling side we should remove that comment.

PR #61, opened 2017-01-26
Try catch chakracore

gabrielschulhof:
Chakracore version of the exception handling API

### DIFF
--- a/src/node_api_helpers.h
+++ b/src/node_api_helpers.h
@@ -14,6 +14,7 @@
 #include "node_asyncapi.h"
 #include <limits.h>
 #include <string.h>
+#include <assert.h>
 
 #define NAPI_METHOD(name)                                                      \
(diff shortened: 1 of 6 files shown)


src/node_api_helpers.h
```diff
@@ -14,6 +14,7 @@
 #include "node_asyncapi.h"
 #include <limits.h>
 #include <string.h>
+#include <assert.h>
```
jasongin, line 17: Why? I don't see where this is used in the added code.

jasongin:
LGTM

src/node_api_helpers.h
gabrielschulhof: It's used below in existing code and, when you include this header, your code fails to build if you don't make this modification.

jasongin: OK, I see.

PR #62, opened 2017-01-26
napi: implement instanceof for ChakraCore

gabrielschulhof:


src/node_api_helpers.h
```diff
@@ -14,6 +14,7 @@
 #include "node_asyncapi.h"
 #include <limits.h>
 #include <string.h>
+#include <assert.h>
```
jasongin, line 17: Is this needed?

jasongin:
LGTM. Thanks for moving to the new branch!

src/node_api_helpers.h
gabrielschulhof: It is, because code below this line makes assertions so, if you include this header without this modification, the resulting code fails to build.

PR #66, opened 2017-01-27
Fix incorrect use of scope term (V8)

### LINKED ISSUE #64: napi_call_function scope parameter
NAPI's call-function API is defined like this:

```C++
napi_value napi_call_function(napi_env e, napi_value scope,
                              napi_value func, int argc, napi_value* argv);
```

What is that scope parameter supposed to be? Both the V8 and JSRT implementations treat it as the `this` argument for the function call, which makes sense except for the `scope` parameter name. Is the parameter mis-named?

This name confusion apparently has caused a problem in the function test case, which [passes a `napi_escapable_handle_scope` as the scope parameter after casting it to a `napi_value`](https://github.com/nodejs/abi-stable-node/blob/api-prototype-8.x/test/addons-abi/test_function/test_function.cc#L21-L24). When running the test on a NAPI V8 branch, the test passes, I assume because the function itself does not reference its `this` arg. But on the chakracore branch, the JSRT function call tries to validate that the parameter is an Object and returns an error code, so the function call fails.

jasongin:
Fixes #64 for the V8 8.x branch. (Once this lands I'll also merge into the V8 6.x branch.)
For the chakracore version of this fix see #65.

### DIFF
--- a/src/node_api_helpers.h
+++ b/src/node_api_helpers.h
@@ -218,7 +218,7 @@ namespace Napi {
     inline napi_value
     Call(int argc, napi_value argv[]) const {
       napi_env env = napi_get_current_env();
-      return Call_(napi_get_global_scope(env), argc, argv);
+      return Call_(napi_get_global(env), argc, argv);
     }
 
    private:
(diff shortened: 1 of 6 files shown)


gabrielschulhof (approved):

mhdawson (approved):
LGTM

jasongin:
Merged into the api-prototype-6.2.0 branch as 5e8ad92ed2fd4f486c803e4c6f44c3197cab2320. I'll port to the api-prototype-8.x branch shortly along with the other recent changes.

PR #68, opened 2017-01-28
Property accessors (ChakraCore)

jasongin:
This is the ChakraCore implementation of property accessors that was already done for V8.

Note all the test case updates as well as `node_jsvmapi.h` and `node_jsvmapi.cc` are just copied from the V8 branch. (The latter is not currently compiled, but I assume we want to keep it around and in sync to someday enable building with either V8 or CC out of the same branch.) The main work in this PR is in `node_jsrtapi.cc`.


src/node_jsrtapi.cc
```diff
@@ -265,60 +179,51 @@ _Ret_maybenull_ JsValueRef CALLBACK CallbackWrapper(JsValueRef callee, bool isCo
...
 
-napi_value napi_create_constructor_for_wrap_with_methods(
-  napi_env e,
-  napi_callback cb,
-  char* utf8name,
-  int methodcount,
-  napi_method_descriptor* methods) {
+napi_value napi_create_constructor(napi_env e, const char* utf8name, napi_callback cb,
+  void* data, int property_count, napi_property_descriptor* properties) {
+  // TODO: set properties
```
boingoing, line 207: Is this still a TODO?

boingoing (approved):
Looks good to me 👍

boingoing:
> The latter is not currently compiled, but I assume we want to keep it around and in sync to someday enable building with either V8 or CC out of the same branch

I think this is a good idea and probably a goal at least for the chakracore branch. There are probably some good candidates for abstracting the two implementations to remove duplication but that is a different effort.

src/node_jsrtapi.cc
jasongin: The latest code in the V8 branches doesn't have this comment. The version I copied here is just a little behind since it was from last week when I created this PR. But it doesn't really matter since this file isn't compiled in this branch.

Issue #69, opened 2017-01-30
../src/node_jsvmapi.cc:192:3: error: ‘interface’ does not name a type


gabrielschulhof:
The error handling modifications fail to build with GCC.

gabrielschulhof:
@boingoing ^

boingoing:
This line is:
```c++
interface CallbackWrapper {
...
```
Does the 'interface' keyword not exist in GCC? Is 'interface' an ISO standard keyword? I'm actually not sure. Looks like it might be an MSVC-specific #define. Maybe we should replace this with class or struct for compliance sake?

gabrielschulhof:
I think so. It certainly doesn't build out of the box.

boingoing:
@jasongin 
Seems like replacing 'interface' with 'struct' should solve the issue. Is there a reason to use 'interface'? I'm actually not super familiar with it.

boingoing:
There is another build issue on gcc, too, so I started a PR. https://github.com/nodejs/abi-stable-node/pull/70

jasongin:
Yes, `interface` is just an alias for `struct`, supported by MSVC but not GCC. I had a habit of using `interface` from Windows programming, since it explicitly declares the intent for the struct to only have abstract methods. But it's certainly not required.

jasongin:
Fixed by https://github.com/nodejs/abi-stable-node/commit/5ae919aa7e040ea1238605778ce8a5f33a91a480

PR #70, opened 2017-01-30
Fix two build issues when building with GCC

boingoing:
First, GCC doesn't seem to implement an interface keyword. Workaround here is to use struct instead of interface. This is https://github.com/nodejs/abi-stable-node/issues/69

Second, GCC doesn't support the flavor of static_assert which takes a single parameter. Just added a second parameter message here.


gabrielschulhof (approved):
LGTM

jasongin:
LGTM. I also see the error about the `static_assert` when building on Windows. So I'm going to go ahead and merge this to unblock me.

PR #72, opened 2017-01-30
Error handling tests and fixes (V8)

jasongin:
- Update tests for EH API changes
 - Update api helpers for EH API changes
 - Fix bugs in EH implementation discovered via testing

All but four of the test cases are updated. I have to leave early today so I'm putting this up for review now. I'll do the rest before this lands, but I expect the patterns to be similar.

In `node_api_helpers.h` I have updated the code so that it builds and works, however return status is currently ignored everywhere. We need to discuss how the helper should handle errors: C++ exceptions, or... ?


boingoing (approved):
Looks good to me. 

Ignoring most of the return codes is probably fine for these tests. If there are any tests which should fail if some operation fails, we should check those. Maybe just asserting the return value?

```c++
assert(napi_get_undefined(e, &result) == napi_ok);
```

jasongin:
I don't think an `assert()` in the C++ code will do the right thing for the test cases that are expecting JS assertions. I considered throwing a JS exception (probably via a macro) whenever an error code was detected, but it didn't seem necessary for what the tests were trying to do. I'm not against it though.

boingoing:
I'm in agreement.

PR #73, opened 2017-01-31
Add a check for pending exception before returning from public API

boingoing:
If there is a pending exception, the API returns napi_pending_exception
and otherwise will return napi_ok.

### DIFF
--- a/src/node_jsvmapi.cc
+++ b/src/node_jsvmapi.cc
@@ -1290,7 +1290,7 @@ napi_status napi_throw(napi_env e, napi_value error) {
       v8impl::V8LocalValueFromJsValue(error));
   // any VM calls after this point and before returning
   // to the javascript invoker will fail
-  return GET_RETURN_STATUS();
+  return napi_ok;
 }
 
 napi_status napi_throw_error(napi_env e, const char* msg) {
@@ -1303,7 +1303,7 @@ napi_status napi_throw_error(napi_env e, const char* msg) {
   isolate->ThrowException(v8::Exception::Error(str));
   // any VM calls after this point and before returning
   // to the javascript invoker will fail
-  return GET_RETURN_STATUS();
+  return napi_ok;
 }
 
 napi_status napi_throw_type_error(napi_env e, const char* msg) {
@@ -1316,7 +1316,7 @@ napi_status napi_throw_type_error(napi_env e, const char* msg) {
   isolate->ThrowException(v8::Exception::TypeError(str));
   // any VM calls after this point and before returning
   // to the javascript invoker will fail
-  return GET_RETURN_STATUS();
+  return napi_ok;
 }
 
 napi_status napi_get_number_from_value(napi_env e, napi_value v, double* result) {
--- a/src/node_jsvmapi.cc
+++ b/src/node_jsvmapi.cc
@@ -1758,7 +1758,7 @@ napi_status napi_is_exception_pending(napi_env e, bool* result) {
   CHECK_ARG(result);
 
   *result = !v8impl::TryCatch::lastException().IsEmpty();
-  return GET_RETURN_STATUS();
+  return napi_ok;
 }
 
 napi_status napi_get_and_clear_last_exception(napi_env e, napi_value* result) {
@@ -1776,7 +1776,7 @@ napi_status napi_get_and_clear_last_exception(napi_env e, napi_value* result) {
     v8impl::TryCatch::lastException().Reset();
   }
 
-  return GET_RETURN_STATUS();
+  return napi_ok;
 }
 
 napi_status napi_buffer_new(napi_env e, char* data, uint32_t size, napi_value* result) {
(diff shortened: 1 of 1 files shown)


src/node_jsvmapi.cc
```diff
@@ -514,6 +514,10 @@ namespace v8impl {
 #define CHECK_NEW_FROM_UTF8(isolate, result, str)                       \
     CHECK_NEW_FROM_UTF8_LEN((isolate), (result), (str), -1)
 
+#define GET_RETURN_STATUS()                                             \
+  v8impl::TryCatch::lastException().IsEmpty()                           \
```
jasongin: This won't work, because the static `TryCatch::lastException()` doesn't get set until the `TryCatch` instance on the stack gets destructed, which happens _after_ the return statement. Instead, check `tryCatch.HasCaught()`.

src/node_jsvmapi.cc
```diff
@@ -514,6 +514,10 @@ namespace v8impl {
...
+    ? napi_ok : napi_pending_exception
```
jasongin: Wrap the whole macro expansion in parentheses in case it is used in a larger expression.

src/node_jsvmapi.cc
boingoing: Good point. Problem with this: tryCatch in NAPI_PREAMBLE is defined inside the do {} while(0). Sorry, didn't catch this when refactoring that macro. I propose removing the do {} while(0) which un-breaks the tryCatch local. Don't think it makes sense to provide a separate macro or put only some of the lines of NAPI_PREAMBLE inside a scope.

src/node_jsvmapi.cc
boingoing: This is a good catch. Thanks.

src/node_jsvmapi.cc
```diff
@@ -514,6 +512,9 @@ namespace v8impl {
 #define CHECK_NEW_FROM_UTF8(isolate, result, str)                       \
     CHECK_NEW_FROM_UTF8_LEN((isolate), (result), (str), -1)
 
+#define GET_RETURN_STATUS()                                             \
+  (tryCatch.HasCaught() ? napi_ok : napi_pending_exception)
+
```
mhdawson: I' thinking this should use napi_set_last_error() like the other macros do.

src/node_jsvmapi.cc
```diff
@@ -481,13 +481,11 @@ namespace v8impl {
...
-    RETURN_STATUS_IF_FALSE(v8impl::TryCatch::lastException().IsEmpty(), \
-      napi_pending_exception);                                          \
-    napi_clear_last_error();                                            \
-    v8impl::TryCatch tryCatch(v8impl::V8IsolateFromJsEnv((e)));         \
-  } while(0)
+  CHECK_ARG(e);                                                         \
+  RETURN_STATUS_IF_FALSE(v8impl::TryCatch::lastException().IsEmpty(),   \
+    napi_pending_exception);                                            \
+  napi_clear_last_error();                                              \
+  v8impl::TryCatch tryCatch(v8impl::V8IsolateFromJsEnv((e)))
```
mhdawson: I see you removed the do while here.  Do we still need it in the other places ?

mhdawson (commented):
Overall looks reasonable, just a few comments.

src/node_jsvmapi.cc
jasongin: The do-while had to be removed because the `tryCatch` must have function scope. I made the same fix in my PR https://github.com/nodejs/abi-stable-node/pull/72 after the problem was revealed by test failures.

The other macros may continue to use do-while because it makes them behave more like a single statement rather than a statement group; as you're probably aware that's a common pattern. It's not strictly necessary just for our internal use though.

src/node_jsvmapi.cc
jasongin: +1

jasongin: Also the condition is backwards.

boingoing: Good catch on the condition. I'll add a call to napi_set_last_error only in the pending exception case. The stored error code should still be napi_ok if there was no previous error.

src/node_jsvmapi.cc
jasongin: I merged https://github.com/nodejs/abi-stable-node/pull/72, so you'll need to rebase on top of that.

Then you can run `vcbuild release nosign test-addons-abi` to run all the ABI tests, which should be all passing now in this branch.

src/node_jsvmapi.cc
```diff
@@ -1289,7 +1290,7 @@ napi_status napi_throw(napi_env e, napi_value error) {
       v8impl::V8LocalValueFromJsValue(error));
   // any VM calls after this point and before returning
   // to the javascript invoker will fail
-  return napi_ok;
+  return GET_RETURN_STATUS();
```
jasongin: Should this return `napi_ok`? As it is here, it will return `napi_pending_exception`, which is also correct I guess. But I think I'd expect a function that throws an exception to return a success code when the exception is successfully thrown.

Same for the following two functions.

boingoing: Yeah, good point. I agree this should just return a flat napi_ok from here. I guess we could check to see if the result of ThrowException is empty and return an error in that case.

jasongin:
LGTM

Issue #63, opened 2017-01-26
Bugs in napi_get_string_from_value


jasongin:
I found multiple problems with the current implementations of `napi_get_string_from_value`: 
 * The function is currently defined as returning the number of bytes _remaining_ in 
the buffer. This is different from both [V8](https://v8docs.nodesource.com/node-0.8/d2/db3/classv8_1_1_string.html#a237a2b794f31d8c60d9561ad3b30248e) and [JSRT](https://github.com/Microsoft/ChakraCore/wiki/JsCopyStringUtf8) string APIs, which always return the number of bytes 
_written_. I think this behavior will definitely confuse some people and cause bugs, so it should be fixed. 
 * In the part of the V8 implementation that copies a stringified number value, the null-terminator byte is 
not copied: it just does a `memcpy` with the length of the string, where the length does not include the 
terminator. 
 * The JSRT implementation does not handle numbers like the V8 implementation does. 
 * If it's possible to get a string from a number value, why not other primitive types (symbol, boolean, 
null/undefined)? We should check the V8 behavior for all of those and try to be consistent. 
 * The JSRT implementation does not copy the null terminator. It assumes `JsCopyStringUtf8` does copy 
the terminator (it does not), and that incorrect assumption also leads to an incorrect 
remaining return value. 

That last item in the list above causes the `test_string` and `test_object` test cases to fail in the NAPI 
chakracore branch.

jasongin:
The `Utf8String` constructor in `node_api_helpers.h` also mistakenly assumes the return value from `napi_get_string_value` is the number of bytes written, not remaining.

bnoordhuis:
> If it's possible to get a string from a number value, why not other primitive types (symbol, boolean, null/undefined)? We should check the V8 behavior for all of those

It's quite complex.  `v8::Utf8Value::Utf8Value(...`):

1. Performs the equivalent of .toString() on primitives (strings, numbers, etc, but not symbols.)
2. Tries `[Symbol.toPrimitive]`, .toString() and .valueOf() in that order on objects.
3. If those methods don't exist, don't complete successfully, or don't produce a string, it raises an exception.
4. Trying to stringify a symbol raises an exception.

I say "equivalent of .toString()" because it doesn't actually invoke the method.  Monkey-patching e.g. `Number.prototype.toString` has no effect.

PR #83, opened 2017-02-02
String API improvements (V8)

jasongin:
- Make string API names consistent
 - Make string API coercion behavior consistent
 - Switch from number-remaining to number-copied results
 - Add UTF-16 string APIs
 - Update tests for API changes

Fixes https://github.com/nodejs/abi-stable-node/issues/63


src/node_jsvmapi.cc
```diff
@@ -1002,28 +1002,29 @@ napi_status napi_create_array_with_length(napi_env e, int length, napi_value* re
   return GET_RETURN_STATUS();
 }
 
-napi_status napi_create_string(napi_env e, const char* s, napi_value* result) {
+napi_status napi_create_string_utf8(napi_env e, const char* s, size_t length,
```
jasongin: Oh, I need to either change `size_t length` back to `int` or restore the without-length variant for creating a string value. My intention was that length here could be -1 when the length is not known by the caller, except `size_t` is unsigned.

sampsongao: Probably restoring to int make more sense, since V8 NewFromUtf8 is taking int with default value to -1 as well.

src/node_jsvmapi.cc
```diff
@@ -1407,34 +1365,98 @@ napi_status napi_get_value_bool(napi_env e, napi_value v, bool* result) {
...
+// insufficient, the string will be truncated, including a null terminator.
+napi_status napi_get_value_string_utf8(
+    napi_env e, napi_value v, char* buf, size_t bufsize, size_t* result) {
+  NAPI_PREAMBLE(e);
+
+  v8::Isolate *isolate = v8impl::V8IsolateFromJsEnv(e);
+  v8::Local<v8::Context> context = isolate->GetCurrentContext();
+  v8::Local<v8::String> str;
+
+  CHECK_TO_STRING(context, str, v);
```
jasongin: Now I'm thinking maybe these `napi_get_value_*` functions should not automatically try to *coerce* the value to a string. Rather, there should be a helper method in the C++ wrapper to do that if desired. Something like `ToString()` vs `AsString()`.

src/node_jsvmapi.cc
jasongin: Changed back to `int`, and made `int` vs `size_t` usage more consistent everywhere.

src/node_jsvmapi.cc
```diff
@@ -1319,122 +1318,143 @@ napi_status napi_throw_type_error(napi_env e, const char* msg) {
...
   CHECK_ARG(result);
 
-  *result = v8impl::V8LocalValueFromJsValue(v).As<v8::String>()
-      ->WriteUtf8(
-          buf,
-          bufsize,
-          0,
-          v8::String::NO_NULL_TERMINATION | v8::String::REPLACE_INVALID_UTF8);
+  v8::Local<v8::Value> value = v8impl::V8LocalValueFromJsValue(v);
+  if (!value->IsString()) return napi_set_last_error(napi_string_expected);
```
boingoing: This pattern could be replaced by the RETURN_STATUS_IF_FALSE macro.

jasongin: Yes, good point. I'll do that.

src/node_jsvmapi.cc
```diff
@@ -1319,122 +1318,143 @@ napi_status napi_throw_type_error(napi_env e, const char* msg) {
...
 }
 
-napi_status napi_get_value_int64(napi_env e, napi_value v, int64_t* result) {
+napi_status napi_get_value_bool(napi_env e, napi_value v, bool* result) {
   NAPI_PREAMBLE(e);
   CHECK_ARG(result);
 
-  *result = v8impl::V8LocalValueFromJsValue(v)->IntegerValue();
+  v8::Local<v8::Value> value = v8impl::V8LocalValueFromJsValue(v);
+  if (!value->IsBoolean()) return napi_set_last_error(napi_boolean_expected);
```
boingoing: These napi_get_value_* API don't do coercion anymore, so what's the pattern we think integrators should use?

```c++
bool r;
napi_status s = napi_get_value_bool(env, possibly_bool, &r);
if (s == napi_boolean_expected) {
  // ?
} else if (s == napi_ok) {
  return r;
} else {
  CHECK_STATUS(napi_throw_error(env, "Boolean expected!"));
}
```

Or do we provide a ```napi_coerce_to_bool``` function?

```c++
bool r;
napi_value definitely_bool;
CHECK_STATUS(napi_coerce_to_bool(env, possibly_bool, &definitely_bool));
CHECK_STATUS(napi_get_value_bool(env, definitely_bool, &r);
return r;
```

Or push the logic into module?
```c++
bool r;
napi_valuetype t;
CHECK_STATUS(napi_get_type_of_value(env, possibly_bool, &t));
switch (t) {
case napi_boolean:
  CHECK_STATUS(napi_get_value_bool(env, possibly_bool, &r);
  return r;
case napi_number:
  int32_t temp;
  CHECK_STATUS(napi_get_value_int32(env, possibly_bool, &temp));
  return (bool)temp;
...
}
```

jasongin: As we discussed, since coercion is no longer automatic we need to add at least the following APIs:
 * `napi_coerce_to_bool`
 * `napi_coerce_to_number`

As a later optimization, we may consider a new set of APIs for getting values with automatic coercion, e.g. `napi_get_value_to_bool`. But we don't need to do that now.

src/node_jsvmapi.cc
```diff
@@ -495,6 +495,12 @@ namespace v8impl {
 #define CHECK_TO_STRING(context, result, src)                           \
     CHECK_TO_TYPE(String, context, result, src, napi_string_expected)
 
+#define CHECK_TO_NUMBER(context, result, src)                           \
+    CHECK_TO_TYPE(Number, context, result, src, napi_string_expected)
+
+#define CHECK_TO_BOOL(context, result, src)                             \
+    CHECK_TO_TYPE(Boolean, context, result, src, napi_string_expected)
```
boingoing: Is napi_string_expected right here?

jasongin: Obviously not! You caught a copy-paste error.

src/node_jsvmapi.cc
```diff
@@ -499,7 +499,7 @@ namespace v8impl {
     CHECK_TO_TYPE(Number, context, result, src, napi_string_expected)
```
boingoing: Apologies I didn't add a separate comment earlier, but I believe this one is also a copy/paste mistake.

boingoing:
👍

PR #84, opened 2017-02-02
String API improvements (ChakraCore)

jasongin:
- Make string API names consistent
 - Make string API coercion behavior consistent
 - Switch from number-remaining to number-copied results
 - Ensure null-terminator is written
 - Add UTF-16 string APIs
 - Update tests for API changes
 - Fix test that incorrectly casted a napi_value to napi_propertyname (worked with V8, not CC)

Fixes https://github.com/nodejs/abi-stable-node/issues/63

With these fixes, all `test/addons-abi` test cases are now passing in the chakracore branch (as well as V8).


boingoing:
LGTM 🚢

PR #88, opened 2017-02-06
C++ wrapper classes

jasongin:
I've filled out and refined the concept I showed last week. The object model is mostly complete now; I think the only missing piece is corresponding wrappers for the libuv async APIs.

I don't think you'll find anything too surprising here since it's mostly a faithful object-oriented interpretation of the NAPI APIs we already had. The interesting/clever parts to look at are:
 * `Napi::Value` class hierarchy and `As*` / `To*` conversions. (Actually this is very similar to V8's object model.)
 * JavaScript & C++ exception unification: see comments on the `Napi::Error` class.
 * `ObjectWrap` callback registration and handling that enables natural use of `this`, return values, and exceptions.

Overall, I think the result is a better/cleaner development experience than what you get from using NAN or Node/V8 APIs directly. Hopefully that will be another reason native module developers might want to use this API.

I kept the type declarations (`napi.h`) separate from the implementations (`napi-inl.h`) because it's much easier get a grasp of the object model that way, at least in the absence of any other documentation. I think the new headers should completely supersede `node_api_helpers.h`, though I have not modified or deleted that file yet because some test/example code still uses it.

I also made a couple small changes to the C API surface: `napi_define_class` and `napi_define_properties`, to better support / be more consistent with the C++ wrappers.

I validated all this using the canvas module, which is now fully ported using these C++ APIs, and covers a very significant portion of them. The Canvas tests all pass, with the exception of those that are dependent on typed arrays. For examples of converted code see [`Canvas.h`](https://github.com/jasongin/node-canvas/blob/napi2/src/Canvas.h) and [`Canvas.cc`](https://github.com/jasongin/node-canvas/blob/napi2/src/Canvas.cc) or the [complete diff of the Canvas port to NAPI](https://github.com/Automattic/node-canvas/compare/master...jasongin:napi2?expand=1).

While I've included the new headers in the regular node `src` directory for now, I actually think they might belong in their own repo, where they'd get published via an NPM, similar to NAN.

### DIFF
--- a/src/napi-inl.h
+++ b/src/napi-inl.h
@@ -1,0 +1,1484 @@
+﻿#ifndef SRC_NAPI_INL_H_
+#define SRC_NAPI_INL_H_
+
+////////////////////////////////////////////////////////////////////////////////
+// NAPI C++ Wrapper Classes
+//
+// Inline header-only implementations for "NAPI" ABI-stable C APIs for Node.js.
+////////////////////////////////////////////////////////////////////////////////
+
+#include <cassert>
+
+namespace Napi {
+
+// Adapt the NODE_MODULE registration function to NAPI:
+//  - Wrap the arguments in NAPI wrappers.
+//  - Catch any NAPI errors that might be thrown.
+#define NAPI_MODULE(modname, regfunc)                                                 \
+  void __napi_ ## regfunc(napi_env env, napi_value exports, napi_value module) {      \
+    try {                                                                             \
+      regfunc(Napi::Env(env), Napi::Object(env, exports), Napi::Object(env, module)); \
+    }                                                                                 \
+    catch (const Napi::Error&) {                                                      \
+      assert(false); /* Uncaught error in native module registration. */              \
+    }                                                                                 \
+  }                                                                                   \
+  NODE_MODULE_ABI(modname, __napi_ ## regfunc);
+
+
+////////////////////////////////////////////////////////////////////////////////
+// Env class
+////////////////////////////////////////////////////////////////////////////////
+
+inline Env::Env(napi_env env) : _env(env) {
+}
+
+inline Env::operator napi_env() const {
+  return _env;
+}
+inline Env Env::Current() {
+  napi_env env;
+  napi_status status = napi_get_current_env(&env);
+  if (status != napi_ok) env = nullptr; // Can't throw without an environment!
+  return Env(env);
+}
+
+inline Object Env::Global() const {
+  napi_value value;
+  napi_status status = napi_get_global(*this, &value);
+  if (status != napi_ok) throw const_cast<Env*>(this)->NewError();
+  return Object(*this, value);
+}
+
+inline Value Env::Undefined() const {
+  napi_value value;
+  napi_status status = napi_get_undefined(*this, &value);
+  if (status != napi_ok) throw const_cast<Env*>(this)->NewError();
+  return Value(*this, value);
+}
+
+inline Value Env::Null() const {
+  napi_value value;
+  napi_status status = napi_get_null(*this, &value);
+  if (status != napi_ok) throw const_cast<Env*>(this)->NewError();
+  return Value(*this, value);
+}
+
+inline Error Env::NewError() {
+  napi_value error;
+  if (IsExceptionPending()) {
+    napi_get_and_clear_last_exception(_env, &error);
+  }
+  else {
+    // No JS exception is pending, so check for NAPI error info.
+    const napi_extended_error_info* info = napi_get_last_error_info();
+
+    napi_value message;
+    napi_status status = napi_create_string_utf8(
+      _env,
+      info->error_message != nullptr ? info->error_message : "Error in native callback",
+      -1,
+      &message);
+    assert(status == napi_ok);
+
+    if (status == napi_ok) {
+      switch (info->error_code) {
+        case napi_object_expected:
+        case napi_string_expected:
+        case napi_boolean_expected:
+        case napi_number_expected:
+          status = napi_create_type_error(_env, message, &error);
+          break;
+        default:
+          status = napi_create_error(_env, message, &error);
+          break;
+      }
+      assert(status == napi_ok);
+    }
+  }
+
+  return Error(_env, error);
+}
+
+inline Error Env::NewError(const char* message) {
+  napi_value str;
+  napi_status status = napi_create_string_utf8(*this, message, -1, &str);
+  if (status != napi_ok) throw NewError();
+
+  napi_value error;
+  status = napi_create_error(*this, str, &error);
+  if (status != napi_ok) throw NewError();
+
+  return Error(*this, error);
+}
+
+inline Error Env::NewTypeError(const char* message) {
+  napi_value str;
+  napi_status status = napi_create_string_utf8(*this, message, -1, &str);
+  if (status != napi_ok) throw NewError();
+
+  napi_value error;
+  status = napi_create_error(*this, str, &error);
+  if (status != napi_ok) throw NewError();
+
+  return Error(*this, error);
+}
+
+inline Buffer Env::NewBuffer(char* data, size_t size) {
+  napi_value value;
+  napi_status status = napi_buffer_new(*this, data, size, &value);
+  if (status != napi_ok) throw NewError();
+  return Buffer(*this, value);
+}
+
+inline Buffer Env::CopyBuffer(const char* data, size_t size) {
+  napi_value value;
+  napi_status status = napi_buffer_copy(*this, data, size, &value);
+  if (status != napi_ok) throw NewError();
+  return Buffer(*this, value);
+}
+
+inline Number Env::NewNumber(double val) {
+  napi_value value;
+  napi_status status = napi_create_number(_env, val, &value);
+  if (status != napi_ok) throw NewError();
+  return Number(_env, value);
+}
+
+inline Boolean Env::NewBoolean(bool val) {
+  napi_value value;
+  napi_status status = napi_create_boolean(_env, val, &value);
+  if (status != napi_ok) throw NewError();
+  return Boolean(_env, value);
+}
+
+inline String Env::NewStringUtf8(const char* val, int length) {
+  napi_value value;
+  napi_status status = napi_create_string_utf8(_env, val, length, &value);
+  if (status != napi_ok) throw NewError();
+  return String(_env, value);
+}
+
+inline String Env::NewStringUtf16(const char16_t* val, int length) {
+  napi_value value;
+  napi_status status = napi_create_string_utf16(_env, val, length, &value);
+  if (status != napi_ok) throw NewError();
+  return String(_env, value);
+}
+
+inline Object Env::NewObject() {
+  napi_value value;
+  napi_status status = napi_create_object(_env, &value);
+  if (status != napi_ok) throw NewError();
+  return Object(_env, value);
+}
+
+inline Array Env::NewArray() {
+  napi_value value;
+  napi_status status = napi_create_array(_env, &value);
+  if (status != napi_ok) throw NewError();
+  return Array(_env, value);
+}
+
+inline Array Env::NewArray(int length) {
+  napi_value value;
+  napi_status status = napi_create_array_with_length(_env, length, &value);
+  if (status != napi_ok) throw NewError();
+  return Array(_env, value);
+}
+
+inline Function Env::NewFunction(VoidFunctionCallback cb, const char* utf8name, void* data) {
+  CallbackData* callbackData = new CallbackData({}); // TODO: Delete when the function is destroyed
+  callbackData->voidFunctionCallback = cb;
+  callbackData->data = data;
+
+  // TODO: set the function name
+  napi_value value;
+  napi_status status = napi_create_function(
+    _env, VoidFunctionCallbackWrapper, callbackData, &value);
+  if (status != napi_ok) throw NewError();
+  return Function(_env, value);
+}
+
+inline Function Env::NewFunction(FunctionCallback cb, const char* utf8name, void* data) {
+  CallbackData* callbackData = new CallbackData({}); // TODO: Delete when the function is destroyed
+  callbackData->functionCallback = cb;
+  callbackData->data = data;
+
+  // TODO: set the function name
+  napi_value value;
+  napi_status status = napi_create_function(
+    _env, FunctionCallbackWrapper, callbackData, &value);
+  if (status != napi_ok) throw NewError();
+  return Function(_env, value);
+}
+
+inline void Env::VoidFunctionCallbackWrapper(napi_env env, napi_callback_info info) {
+  napi_value result;
+  try {
+    CallbackInfo callbackInfo(env, info);
+    CallbackData* callbackData = reinterpret_cast<CallbackData*>(callbackInfo.Data());
+    callbackData->functionCallback(callbackInfo);
+  }
+  catch (const Error& e) {
+    if (!Env(env).IsExceptionPending()) {
+      e.ThrowAsJavaScriptException();
+    }
+    return;
+  }
+}
+
+inline void Env::FunctionCallbackWrapper(napi_env env, napi_callback_info info) {
+  napi_value result;
+  try {
+    CallbackInfo callbackInfo(env, info);
+    CallbackData* callbackData = reinterpret_cast<CallbackData*>(callbackInfo.Data());
+    result = callbackData->functionCallback(callbackInfo);
+  }
+  catch (const Error& e) {
+    if (!Env(env).IsExceptionPending()) {
+      e.ThrowAsJavaScriptException();
+    }
+    return;
+  }
+
+  napi_status status = napi_set_return_value(env, info, result);
+  if (status != napi_ok) return;
+}
+
+inline bool Env::IsExceptionPending() const {
+  bool result;
+  napi_status status = napi_is_exception_pending(_env, &result);
+  if (status != napi_ok) result = false; // Checking for a pending exception shouldn't throw.
+  return result;
+}
+
+inline void Env::Throw(Value error) {
+  napi_throw(_env, error);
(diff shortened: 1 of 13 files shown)


jasongin:
This is for issue #53

jasongin:
Forgot to mention: there are several TODOs in the code related to `CallbackData` allocations that never get deleted. Since callbacks are typically only set up once when the module is registered, it's not a major problem. But technically the memory is leaked since even if the function object is GC'd the `CallbackData` allocation is never deleted. I need to investigate the best way to handle that (or maybe someone can make a suggestion?), but I didn't want to block everything else on that.

src/napi-inl.h
```diff
@@ -0,0 +1,1484 @@
...
+}
+inline Env Env::Current() {
+  napi_env env;
+  napi_status status = napi_get_current_env(&env);
+  if (status != napi_ok) env = nullptr; // Can't throw without an environment!
+  return Env(env);
+}
+
+inline Object Env::Global() const {
+  napi_value value;
```
mhdawson: These could probably be cached right ?   Maybe make Global and Undefined static methods that cache the values.  For Current, it could probably be cached as an instance value and if the env is passed through as objects are created then most times we'd avoid the round trip.

mhdawson (commented):
Still need more time to go through this, but I like what I see so far.  

I think it would be useful to have another version of the add-on examples(https://github.com/nodejs/abi-stable-node-addon-examples) where we use the C++ API.   I'm thinking in a new directory called abi-sugar or something like that.  We'd end up with something like  5 dirs:
* abi
* abi-sugar
* nan
* node_0.10
* node_0.12

Which would should how to implement the examples for each of the cases.   It would be a good way to show how the common cases are going to look in the C abi version, with the C++ sugar and be able to compare against nan and direct v8 implementation.

As you pointed out the C++ wrapper should be moved out of node itself to another project as an add-on to make using the abi stable api easier.

src/napi-inl.h
jasongin: I thought about caching the global and undefined values, but I wasn't sure that the pointers would remain valid without going through a persistent (which would defeat the purpose of caching).

The `Env::Current()` static method should be almost never used. The other object instances all hold onto their environment and pass it along to any new objects they return.

jasongin:
> the C++ wrapper should be moved out of node itself to another project

Should we create a new repo under the nodejs org? What should it be named?

src/napi.h
```diff
@@ -0,0 +1,558 @@
...
+    Error NewTypeError(const char* message);
+    Buffer NewBuffer(char* data, size_t size);
+    Buffer CopyBuffer(const char* data, size_t size);
+    Number NewNumber(double val);
+    Boolean NewBoolean(bool val);
+    String NewStringUtf8(const char* val, int length = -1);
+    String NewStringUtf16(const char16_t* val, int length = -1);
+    Array NewArray();
+    Array NewArray(int length);
+    Object NewObject();
```
jasongin: To be more similar to V8 (and make porting easier), I'm going to move all these `New*` methods to static methods on their respective classes. Of course, then they will require an `Env` instance as the first parameter.

mhdawson: ok, if you post a message when you have completed that I'll wait to take another pass until you have done that.

mhdawson:
@jasongin in terms of creating a new repo I'm thinking what we might name it.  How does "abi-stable-node-wrapper" sound ? I'm thinking we'll rename to something better at some point.  For now it would show a clear association with the work in abi-stable-node

src/napi.h
jasongin: It may be helpful to review my typed-array PRs first, since I'm also going to rebase this PR on top of those changes so I can include typed-array support in the wrapper classes.

jasongin:
I agree it makes sense to have all the repos start with the same prefix.

Longer term, when "abi-stable-node" is just some APIs exposed by node core, I think the wrapper component needs a better name.

jasongin:
I rebased on top of the typed-arrays changes, and added another commit:
 - Moved static `New` methods from `Napi::Env` to their respective `Napi::Value` subclasses.
 - Added typed-array classes in wrapper.
 - Added RangeError support in both the C API and the wrapper.

With typed-arrays support, I was able to complete the node-canvas port. All the canvas tests pass! (I will try benchmarking it next.)

src/napi.h
```diff
@@ -0,0 +1,663 @@
...
+    std::string Message() const;
+    void ThrowAsJavaScriptException() const;
+
+    const char* what() const override;
+
+  private:
+    std::string _message;
+  };
+
+  class Persistent {
```
jasongin: I am updating this to `Persistent<T>`, where `T` may be any `Value` subclass, and the `Value()` method returns a `T` instead of a `Value`.

src/napi.h
```diff
@@ -0,0 +1,663 @@
...
+  public:
+    Persistent();
+    Persistent(napi_env env, napi_persistent persistent);
+    ~Persistent();
+
+    // A persistent instance can be moved but cannot be copied.
```
jasongin: Probably I should have the copy-constructor / assignment operator create a new persistent reference instead.

src/node_jsvmapi_types.h
```diff
@@ -19,7 +21,10 @@ enum napi_property_attributes {
   napi_default = 0,
   napi_read_only = 1 << 0,
   napi_dont_enum = 1 << 1,
-  napi_dont_delete = 1 << 2
+  napi_dont_delete = 1 << 2,
+
+  // For internal use only by the NAPI C++ wrapper.
+  napi_instance_property = 1 << 10,
```
mhdawson: Should there be anything in the napi c api which is linked to the C++ wrapper, the C++ wrapper should be completely independent.

mhdawson:
Its lots of code to go through but I don't see anything that concerns me yet.  I think it would be worth you walking us through the highlights in our next meeting (for example the exception class).

I'll create new new repo so you can submit a PR to put the wrapper there, we need to include a REAMDE.md md that explains the purpose/linkage to the abi-stable-node repo.

mhdawson:
Repo created: https://github.com/nodejs/abi-stable-node-wrapper

src/node_jsvmapi_types.h
jasongin: Agreed. I didn't like doing this at the time.

Now, I'm actually thinking of keeping this as part of the public C API (actually switching it to `napi_static_property`) and adding support for it in `napi_define_class()` (formerly `napi_create_constructor()`). The purpose is to allow having a single array of `napi_property_descriptor`s that are passed to a single call to `napi_define_class()`, for defining both instance and static properties & methods.

mhdawson:
For the TODO's which I see I think both in the canvas port and the wrapper would it make sense to pull these out into an issue so that we don't forget the list that we have to go back and address ?

mhdawson:
Would it make sense to covert to 2 pull requests.  1) for the changes that will go into node and then a second one as the initial commit of the wrapper into the abi-stable-node-wrapper repo ?

src/node_jsvmapi.cc
```diff
@@ -1881,3 +1881,187 @@ napi_status napi_buffer_length(napi_env e, napi_value v, size_t* result) {
...
+  v8::ArrayBuffer::Contents contents = value.As<v8::ArrayBuffer>()->GetContents();
+
+  if (data != nullptr) {
+    *data = contents.Data();
+  }
+
+  if (byte_length != nullptr) {
+    *byte_length = contents.ByteLength();
+  }
+
```
mhdawson: Should we return an error if both the data and byte_length values are nullptr ?

src/napi.h
kkoopa: Unless I am missing some important detail, that sounds very dangerous. The reason V8 redid `Persistent`s in the version around Node 0.11 was to avoid the problem of careless leaks. The old `Persistent` API found in Node 0.10 allowed implicit copying of `Persistent` handles which led to a whole bunch of leaks due to misuse. It should be impossible to accidentally copy a `Persistent` handle.

src/napi-inl.h
kkoopa: If I remember correctly, regarding V8, the primitives are not handled as normal pointers or such. They are special values of sorts:
```c++
Local<Primitive> Null(Isolate* isolate) {
  typedef internal::Object* S;
  typedef internal::Internals I;
  I::CheckInitialized(isolate);
  S* slot = I::GetRoot(isolate, I::kNullValueRootIndex);
  return Local<Primitive>(reinterpret_cast<Primitive*>(slot));
}
```
So as far as V8 is concerned, there should be no problem with caching primitives. I do not know how other engines handle primtives.

src/napi-inl.h
```diff
@@ -0,0 +1,1724 @@
...
+inline Value::operator napi_value() const {
+  return _value;
+}
+
+inline bool Value::operator ==(const Value& other) const {
+  return StrictEquals(other);
+}
+
+inline bool Value::operator !=(const Value& other) const {
+  return !StrictEquals(other);
```
kkoopa: It is preferred to directly define this as negation of the equality operator.

src/napi-inl.h
```diff
@@ -0,0 +1,1724 @@
...
+  return TypedArray(_env, _value);
+}
+
+inline Function Value::AsFunction() const {
+  return Function(_env, _value);
+}
+
+inline Buffer Value::AsBuffer() const {
+  return Buffer(_env, _value);
+}
```
kkoopa: Why did you opt for this instead of a function template `As<T>`?

src/napi-inl.h
```diff
@@ -0,0 +1,1724 @@
...
+  if (status != napi_ok) throw Error::New(Env());
+  return String(_env, result);
+}
+
+inline Object Value::ToObject() const {
+  napi_value result;
+  napi_status status = napi_coerce_to_object(_env, _value, &result);
+  if (status != napi_ok) throw Error::New(Env());
+  return Object(_env, result);
+}
```
kkoopa, line 243: Why did you opt for this instead of a function template `To<T>`?

src/napi.h
```diff
@@ -0,0 +1,663 @@
...
+    void Set(const char* utf8name, bool boolValue);
+    void Set(const char* utf8name, double numberValue);
+
+    void DefineProperty(const PropertyDescriptor& property);
+    void DefineProperties(const std::vector<PropertyDescriptor>& properties);
+    bool InstanceOf(const Function& constructor) const;
+    template<typename T> T* Unwrap() const;
+  };
+
+  class Array : public Value {
```
kkoopa: Why does `Array` not inherit `Object`?

src/napi.h
```diff
@@ -0,0 +1,663 @@
...
+    T* Data();
+    const T* Data() const;
+
+  private:
+    T* _data;
+
+    _TypedArray(napi_env env, napi_value value, size_t length, T* data);
+  };
+
+  class Function : public Value {
```
kkoopa: Why does `Function` not inherit `Object`?

src/napi.h
jasongin: OK, makes sense, if copies are ever really necessary it is better to force them to be explicit.

src/napi.h
jasongin: I specifically chose not to do that. The only thing it would gain by inheriting from `Object` would be all the property get/set/etc methods, which would rarely ever be used on an array. In the rare case someone does need to treat an array as an object, it is trivial to call `.AsObject()` on the array first.

src/napi.h
jasongin: I can see some value in `Function` inheriting from `Object`, since it is not too unusual to access properties on a function. But still without that, if you want to treat a function as an object it is trivial to call `.AsObject()` first. So, I don't feel strongly either way.

src/napi.h
kkoopa: I find it somewhat confusing when the inheritance hierarchy does not correspond to the ES inheritance hierarchy. In JavaScript, `Array`s are `Object`s, it seems logical that the same would hold here.
Inheritance Diagram

src/napi-inl.h
jasongin: No good reason. It would probably be better to use `As<T>` to be more similar to V8.

src/napi-inl.h
jasongin: 1. I'd have to use template specialization to call the different `napi_coerce_*` functions for each implementation, so the code wouldn't be any simpler with templates.
2. This is more similar to V8, which has separate ToSomething() methods for each supported type.

src/napi-inl.h
jasongin: OK

jasongin:
> Would it make sense to covert to 2 pull requests. 1) for the changes that will go into node and then a second one as the initial commit of the wrapper into the abi-stable-node-wrapper repo ?

I can do that. Would you prefer it right away, or should I keep updating this PR for now to continue the feedback + diff history?

src/napi-inl.h
kkoopa: Fair enough, although one more thing should be taken into account: ease of use for the API consumer. However, unless doing it like I did in NAN with `To<T>` serving both as `ToXXX` and `XXXValue`, I agree that this is better as is.

jasongin:
> For the TODO's which I see I think both in the canvas port and the wrapper would it make sense to pull these out into an issue so that we don't forget the list that we have to go back and address ?

Yes, I will be sure to do that for any TODOs that are remaining at the time this is merged. Some of the TODOs are related to / dependent upon [the issue I opened about references](https://github.com/nodejs/abi-stable-node/issues/93).

jasongin:
Regarding PR separation, I do think I should at least pull out the changes in the NAPI C APIs (`node_jsvmapi_*`) to a separate PR, since I think they are being (understandably) ignored in this review since they aren't the focus.

src/napi.h
```diff
@@ -0,0 +1,663 @@
...
+   *       return new Example();
+   *     }
+   *
+   *     Napi::Value GetValue(const Napi::CallbackInfo& info);
+   *     void SetValue(const Napi::CallbackInfo& info, const Napi::Value& value);
+   *     Napi::Value DoSomething(const Napi::CallbackInfo& info);
+   *   }
+   */
+  template <typename T>
+  class ObjectWrap {
```
jasongin: This class needs `Ref()` and `Unref()` methods.

I have implemented those methods locally using a combination of NAPI weak references and persistent references, but it could be done better using some [NAPI referencing improvements I proposed](https://github.com/nodejs/abi-stable-node/issues/93). And then with those changes, I would probably change the `Persistent` class above to `Reference<T>`, and have `ObjectWrap<T>` extend `Reference<Object>` so it would inherit the `Ref()` / `Unref()` methods from it.

kkoopa:
I remembered another thing: Please make `std::string` supported everywhere that strings are expected.

mhdawson:
I agree pulling out into separate PRs makes sense now, we can land the changes to the napi api while we continue to discuss/review the wrapper which is a bigger piece.

jasongin:
I submitted the [range error](https://github.com/nodejs/abi-stable-node/pull/95) and [property/class definition](https://github.com/nodejs/abi-stable-node/pull/97) changes as separate PRs. [The typed-arrays PR](https://github.com/nodejs/abi-stable-node/pull/90) was already separate, and merged.

jasongin:
I pushed another commit with lots of updates. Now there are 5 commits in this PR; the first two are actually another PR (not merged yet) that I rebased on top of because this PR builds on it. The last three commits are the actual iterations of this PR. The only files touched by those 3 commits are the added `napi.h` and `napi-inl.h` files, which will eventually go in the other repo when this PR is signed off. (Do not press the merge button here!)

Changes in this iteration include:
 - Replaced `Napi::Persistent` and `Napi::Callback` with a new `Napi::Reference<T>` class
 - Added new `Napi::External` class
 - Updated `ObjectWrap` to extend `Reference<Object>`
 - Fixed `ObjectWrap` instance destruction
 - Removed `Env::Current()`
 - Switched to `std::function` for `Napi::Function` callbacks, enabling lambdas (not needed for wrapped class methods).
 - Addressed other review feedback.

I have updated the [canvas NAPI port](https://github.com/Automattic/node-canvas/compare/master...jasongin:napi2?expand=1) to use these latest changes, and it all works.

ianwjhalliday:
This is marvelous and more or less exactly what I had originally envisioned when I suggested a C++ helper layer on top of the C APIs.  Thank you @jasongin for the good work here.

mhdawson (approved):
I still have some concerns about memory allocation.  However, I think at this point given that it will land in a different repo from abi-stable-node, that landing it so that we can use it in the sqlite port as well makes sense.  We can then look at the perf issues to see if they are related to allocation or something else. 

We may also want to do a perf compare between direct C port versus using this wrapper.

So just a long way of saying lets land this in the abi-stable-node-wrapper repo and iterate from there unless anybody else has objections.

jasongin:
Agreed, we can always iterate more in the new repo as issues come up.

> I still have some concerns about memory allocation

Anything specific? (Aside from the TODOs about leaking callback data, for which I will open an issue to track. I have some ideas about how to fix that, which won't be simple but should not require any changes to either the C or C++ API.)

kkoopa:
Identifiers with leading underscores followed by capital letters are reserved according to the C++ standard. In general, avoid leading underscores. The simplest fix is moving the leading underscore to the end.
`_identifier -> identifier_`

jasongin:
> Identifiers with leading underscores followed by capital letters are reserved 

OK, I'll fix that.

jasongin:
I added `napi.h` and `napi-inl.h` to the wrappers repo: https://github.com/nodejs/abi-stable-node-wrapper/commit/fe1d9fd5a2fa898f09695faeb40f5b251b162517

Globik:
Where can I look onto the AsyncWorker class? Any tuto? And what asyncWorker class have under the hood? Is there any napi delete reference and napi delete async work in it?

bnoordhuis:
@Globik Can you stop spamming unrelated issues?  Stick to nodejs/help#1055, please.

PR #89, opened 2017-02-06
build: place NAPI-related changes behind a configure flag

gabrielschulhof:
Perhaps we should update the blurb at the [top of src/node_jsvmapi.cc](https://github.com/nodejs/abi-stable-node/blob/api-prototype-8.x/src/node_jsvmapi.cc#L1-L15).

### DIFF
--- a/src/node.cc
+++ b/src/node.cc
@@ -29,7 +29,9 @@
 #include "node_lttng.h"
 #endif
 
+#if defined ENABLE_NAPI
 #include "node_jsvmapi_internal.h"
+#endif
 
 #include "ares.h"
 #include "async-wrap.h"
@@ -97,7 +99,6 @@ typedef int mode_t;
 #elif !defined(_MSC_VER)
 extern char **environ;
 #endif
-
 
 namespace node {
 
@@ -2418,6 +2419,7 @@ void DLOpen(const FunctionCallbackInfo<Value>& args) {
     return;
   }
 
+#ifdef ENABLE_NAPI
 // try loading with node 0.10 manner in case of node 0.10
   struct node_module_old* mod;
   char symbol[1024], *base, *pos;
@@ -2475,6 +2477,7 @@ void DLOpen(const FunctionCallbackInfo<Value>& args) {
     mp->nm_context_register_func = NULL;
     mp->nm_modname = mod->modname;
   }
+#endif /* def ENABLE_NAPI */
 
   if (mp == nullptr) {
     uv_dlclose(&lib);
@@ -2482,9 +2485,13 @@ void DLOpen(const FunctionCallbackInfo<Value>& args) {
     return;
   }
 
+#ifdef ENABLE_NAPI
   bool isNapiModule = mp->nm_version == -1;
 
   if (mp->nm_version != NODE_MODULE_VERSION && !isNapiModule) {
+#else /* !defined ENABLE_NAPI */
+  if (mp->nm_version != NODE_MODULE_VERSION) {
+#endif /* def ENABLE_NAPI */
     char errmsg[1024];
     snprintf(errmsg,
              sizeof(errmsg),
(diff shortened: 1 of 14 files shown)


src/node.cc
```diff
@@ -2418,6 +2419,7 @@ void DLOpen(const FunctionCallbackInfo<Value>& args) {
     return;
   }
 
+#ifdef ENABLE_NAPI
 // try loading with node 0.10 manner in case of node 0.10
```
jasongin: Why is support for old modules conditional on ENABLE_NAPI?

src/node_jsvmapi.h
```diff
@@ -18,6 +18,7 @@
 #ifndef SRC_NODE_JSVMAPI_H_
 #define SRC_NODE_JSVMAPI_H_
 
+#include "node.h"
```
jasongin: Can inclusion of `node.h` be avoided? It would be better if all the node APIs (and V8 APIs included by `node.h`) were not available when building a NAPI module.

src/node_jsvmapi.h
```diff
@@ -35,6 +36,15 @@
 #endif
 
 namespace node {
+
+struct node_module_old {
```
jasongin: Why is this moved here? Can it go in `node_jsvmapi_internal.h` instead?

src/node_jsvmapi.h
mhdawson: It would be better if modules needed to include that separately.

src/node_jsvmapi.h
mhdawson: I think it should be able to be in  node_jsvmapi_internal.h.  The only place the external definition should have been necessary was when building against 0.10.x and in that case it would not have been the "old" version but the version in node.h.

src/node_jsvmapi.h
mhdawson: Having said that I think that all modules will have to include node.h as they need the node_module definition which it contains.

mhdawson: As a follow on we may want to refactor node.h so that is not the case because as Jason says we want to avoid leakage of the v8 apis into modules building against the NAPI api.

src/node.cc
mhdawson: node_module_old is a structure which reflects the structure of node_module in the 0.10.x release.  We only need it if we are going to try and load a module compiled against 0.10.x which only makes sense with NAPI (without NAPI you have to recompile your module to use it with newer node versions).

src/node.cc
```diff
@@ -2475,16 +2477,21 @@ void DLOpen(const FunctionCallbackInfo<Value>& args) {
...
     uv_dlclose(&lib);
     env->ThrowError("Module did not self-register.");
     return;
   }
 
+#ifdef ENABLE_NAPI
   bool isNapiModule = mp->nm_version == -1;
 
   if (mp->nm_version != NODE_MODULE_VERSION && !isNapiModule) {
+#else /* !defined ENABLE_NAPI */
```
mhdawson: This is a case were if we get agreement to enable NAPI by default at compile time, we may need a runtime flag to make sure we don't try to load modules were the version is -1 unless the runtime Flag has enabled NAPI

test/addons-abi/test_object/test_object.cc
```diff
@@ -1,4 +1,4 @@
-﻿#include <node_jsvmapi.h>
+#include <node_jsvmapi.h>
```
mhdawson, line 1: There are few changes like this that I'm not sure what actually changed ?  Is it whitespace ?

src/node_jsvmapi.h
jasongin: We got by without any inclusion of `node.h` before. It works because `node_jsvmapi.h` defines its own `NODE_MODULE_ABI` macro that is used for module registration. I don't think any refactoring of `node.h` is necessary.

test/addons-abi/test_object/test_object.cc
gabrielschulhof:
```diff
-<U+FEFF>#include <node_jsvmapi.h>
+#include <node_jsvmapi.h>
```

src/node_jsvmapi.h
mhdawson: ah right, good.

gabrielschulhof:
To keep a reference to the 0.10 stuff I could point a branch to the commit preceding the one that lands these changes.

mhdawson:
One more thought is that we may want to make the default to "on" in the abi-stable-node repo and then change that to off if necessary when we PR to node core.

gabrielschulhof:
@mhdawson roger that! Can do.

mhdawson (approved):
LGTM

Issue #50, opened 2017-01-19
Typed arrays


jasongin:
NAPI should expose APIs for JS typed arrays.

Of the top native modules that I scanned, `canvas` was the only one that used typed arrays. But there must also be some less popular native modules that need them.

jasongin:
V8 typed arrays PR: https://github.com/nodejs/abi-stable-node/issues/90
ChakraCore typed arrays PR: https://github.com/nodejs/abi-stable-node/pull/91

jasongin:
Done for both V8 and ChakraCore.

PR #90, opened 2017-02-07
Typed arrays (V8)

jasongin:
This adds 7 new APIs to support typed arrays and their underlying array buffer objects:
 * `napi_is_arraybuffer`
 * `napi_create_arraybuffer`
 * `napi_create_external_arraybuffer`
 * `napi_get_arraybuffer_info`
 * `napi_is_typedarray`
 * `napi_create_typedarray`
 * `napi_get_typedarray_info`

(There is no need for read/write APIs because C++ code is expected to use the buffer pointer directly.)

Typed arrays are tested with a new `test_typedarray` project under `test/addons-abi`.

Also in this change, I moved some type declarations from `node_jsvmapi.h` to `node_jsvmapi_types.h`, and removed an unnecessary constructor from `napi_extended_error_info` because it was incompatible with C.

### DIFF
--- a/src/node_jsvmapi_types.h
+++ b/src/node_jsvmapi_types.h
@@ -1,6 +1,8 @@
 ﻿#ifndef SRC_NODE_JSVMAPI_TYPES_H_
 #define SRC_NODE_JSVMAPI_TYPES_H_
 
+#include <stdint.h>
+
 // JSVM API types are all opaque pointers for ABI stability
 // typedef undefined structs instead of void* for compile time type safety
 typedef struct napi_env__ *napi_env;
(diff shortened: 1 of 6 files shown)


jasongin:
Related issue: https://github.com/nodejs/abi-stable-node/issues/50

src/node_jsvmapi_types.h
```diff
@@ -34,4 +36,48 @@ struct napi_property_descriptor {
...
+  napi_null,
+  napi_boolean,
+  napi_number,
+  napi_string,
+  napi_symbol,
+  napi_object,
+  napi_function,
+};
+
+enum napi_typedarray_type {
```
boingoing, line 51: Think it would be valuable to put these into the napi_valuetype enum instead? Then napi_get_type_of_value could return the TypedArray type, as an example of use case. Not sure if useful, though. Module code likely knows when it's expecting TypedArray objects. Hmm.

boingoing:
LGTM

src/node_jsvmapi_types.h
jasongin: I think `napi_valuetype` and `napi_get_type_of_value` are intentionally limited to fundamental JavaScript types. (Technically `napi_function` breaks that rule because a function is also an object, but I think I'm OK with that.) Adding other Object subclasses to the enum would make things confusing: `napi_get_type_of_value(myTypedArray) == napi_object` would be false, even though a typed-array is in fact an object. Perhaps that's why V8 provides lots of Is*SomeType*() methods instead.

boingoing: Yeah. May be worth thinking about exposing some of those V8 methods via NAPI wrappers to avoid problems like this. Instead of using napi_get_type_of_value to check for napi_object, just use IsObject or IsTypedArray or IsUint8Array directly. Shouldn't block your commit here, though.

jasongin: You mean [like this](https://github.com/nodejs/abi-stable-node/pull/88/files#diff-57762ba270aadd394d6acdf9f06841a9R111) ?

boingoing: Pretty much exactly the surface I was thinking about, yes. Could be a bit lower overhead if the C API exposed tiny wrappers around those v8 IsSomeType() methods. Not sure that overhead is worth worrying about, though.
```c++
napi_status napi_is_number(napi_env e, napi_value v, bool* result) {
  NAPI_PREAMBLE(e);
  CHECK_ARG(result);
  *result = v8impl::V8LocalValueFromJsValue(v)->IsNumber();
  return GET_RETURN_STATUS();
}
```

ianwjhalliday: Just to follow up here, yes, napi_valuetype is meant to mirror JavaScript typeof (which is why function is included as its own variant).

Whether it is actually useful to mirror typeof for NAPIs consumers, I am not sure.

PR #95, opened 2017-02-09
RangeError APIs (V8)

jasongin:
I pulled this out of my [C++ wrapper classes](https://github.com/nodejs/abi-stable-node/pull/88) PR, since that was doing too many things in the same PR.

This is needed for the node-canvas port to NAPI, because some of its unit tests expect RangeErrors.


boingoing:
👍

PR #97, opened 2017-02-09
Property & class definition improvements (V8)

jasongin:
- Rename napi_create_constructor to napi_define_class and include support for defining static properties on the class at the same time
 - Rename napi_define_property to napi_define_properties to allow for defining multiple properties at the same time
 - Update tests for the changes

I pulled this out of my [C++ wrapper classes](https://github.com/nodejs/abi-stable-node/pull/88) PR, since that was doing too many things in the same PR.

These changes are not absolutely necessary, but while working on the C++ wrapper and node-canvas NAPI port I realized this would be a better more efficient way to initialize all a native module's many C++ callbacks.

### DIFF
--- a/src/node_jsvmapi_types.h
+++ b/src/node_jsvmapi_types.h
@@ -21,7 +21,11 @@ enum napi_property_attributes {
   napi_default = 0,
   napi_read_only = 1 << 0,
   napi_dont_enum = 1 << 1,
-  napi_dont_delete = 1 << 2
+  napi_dont_delete = 1 << 2,
+
+  // Used with napi_define_class to distinguish static properties
+  // from instance properties. Ignored by napi_define_properties.
+  napi_static_property = 1 << 10,
 };
 
 struct napi_property_descriptor {
--- a/test/addons-abi/8_passing_wrapped/binding.cc
+++ b/test/addons-abi/8_passing_wrapped/binding.cc
@@ -42,12 +42,11 @@ void Init(napi_env env, napi_value exports, napi_value module) {
 
   MyObject::Init(env);
 
-  napi_property_descriptor desc = { "createObject", CreateObject };
-  status = napi_define_property(env, exports, &desc);
-  if (status != napi_ok) return;
-
-  napi_property_descriptor desc2 = { "add", Add };
-  status = napi_define_property(env, exports, &desc2);
+  napi_property_descriptor desc[] = {
+    { "createObject", CreateObject },
+    { "add", Add },
+  };
+  status = napi_define_properties(env, exports, sizeof(desc) / sizeof(*desc), desc);
   if (status != napi_ok) return;
 }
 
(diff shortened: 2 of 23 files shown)


src/node_jsvmapi_types.h
```diff
@@ -21,7 +21,11 @@ enum napi_property_attributes {
   napi_default = 0,
   napi_read_only = 1 << 0,
   napi_dont_enum = 1 << 1,
-  napi_dont_delete = 1 << 2
+  napi_dont_delete = 1 << 2,
+
+  // Used with napi_define_class to distinguish static properties
+  // from instance properties. Ignored by napi_define_properties.
+  napi_static_property = 1 << 10,
```
boingoing, line 28: Any particular reason to make this flag 1<<10?

test/addons-abi/8_passing_wrapped/binding.cc
```diff
@@ -42,12 +42,11 @@ void Init(napi_env env, napi_value exports, napi_value module) {
...
-  status = napi_define_property(env, exports, &desc);
-  if (status != napi_ok) return;
-
-  napi_property_descriptor desc2 = { "add", Add };
-  status = napi_define_property(env, exports, &desc2);
+  napi_property_descriptor desc[] = {
+    { "createObject", CreateObject },
+    { "add", Add },
+  };
+  status = napi_define_properties(env, exports, sizeof(desc) / sizeof(*desc), desc);
```
boingoing, line 49: This is a lot cleaner / nicer. 👍

test/addons-abi/test_instanceof/test_instanceof.cc
```diff
@@ -28,10 +28,9 @@ void Init(napi_env env, napi_value exports, napi_value module) {
     { "doInstanceOf", doInstanceOf },
   };
 
-  for (int i = 0; i < sizeof(descriptors) / sizeof(*descriptors); i++) {
-    status = napi_define_property(env, exports, &descriptors[i]);
-    if (status != napi_ok) return;
-  }
+  status = napi_define_properties(
+    env, exports, sizeof(descriptors) / sizeof(*descriptors), descriptors);
+  if (status != napi_ok) return;
```
boingoing, line 33: Not sure this check of status has the same meaning as it did before this change. Same pattern is used in a few of these tests. No harm, I guess, but also looks like not doing anything.

boingoing (approved):
Looks good 👍

src/node_jsvmapi_types.h
jasongin: I wanted to leave space in case V8 adds new property attributes that we need to support. The lower attributes are literally the same constant values as the corresponding V8 attributes; they don't have to be but it makes conversion simpler.

test/addons-abi/test_instanceof/test_instanceof.cc
jasongin: In all the test code I have kept the consistent pattern of checking the status after a NAPI call, even if there are no statements following. It makes it easier to copy-paste and refactor code without accidentally forgetting the error-checking.

src/node_jsvmapi_types.h
mhdawson: A bit late, but should we not just put it at the far end instead of just <<10 ?

PR #98, opened 2017-02-10
Fix build errors

jasongin:
This fixes a couple build errors that I saw on a Mac, after my recent changes that I had built only on Windows. Sorry about that.

(For anyone unfamiliar with the template disambiguator syntax used here, see the bottom of [this page](http://en.cppreference.com/w/cpp/language/dependent_name).)

### DIFF
--- a/src/node_jsvmapi.cc
+++ b/src/node_jsvmapi.cc
@@ -217,7 +217,7 @@ namespace v8impl {
       }
 
       virtual void* Data() override {
-        return _cbdata->GetInternalField(kDataIndex).As<v8::External>()->Value();
+        return v8::Local<v8::External>::Cast(_cbdata->GetInternalField(kDataIndex))->Value();
       }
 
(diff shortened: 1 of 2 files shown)


src/node_jsvmapi.cc
```diff
@@ -217,15 +217,15 @@ namespace v8impl {
       }
 
       virtual void* Data() override {
-        return _cbdata->GetInternalField(kDataIndex).As<v8::External>()->Value();
+        return _cbdata->GetInternalField(kDataIndex).template As<v8::External>()->Value();
```
boingoing: Should this use Cast like v8::Local<v8::External>::Cast() instead of the .template ?

jasongin: OK, I'll change to `Cast()`, since it does the same thing as far as I know.

boingoing:
👍

Issue #93, opened 2017-02-08
Weak/strong reference improvements


jasongin:
The NAPI APIs for referencing values currently consist of the following:
```C
napi_status napi_create_weakref(napi_env e, napi_value v, napi_weakref* result);
napi_status napi_release_weakref(napi_env e, napi_weakref w, napi_value* result);
napi_status napi_get_weakref_value(api_env e, napi_weakref w);

napi_status napi_create_persistent(napi_env e, napi_value v, napi_persistent* result);
napi_status napi_release_persistent(napi_env e, napi_persistent p);
napi_status napi_get_persistent_value(napi_env e, napi_persistent p, napi_value* result);
```

These APIs are missing two important capabilities:
 1. The ability to *promote* a weak reference to a strong reference, or *demote* a strong reference to a weak reference. (Currently this is possible in a roundabout way by retrieving the value and creating the other kind of reference, but it's very inconvenient.)
 2. The ability to attach a finalization callback to any (weak) reference, so that a custom cleanup procedure can be performed if/when the value gets GC'd. (Currently NAPI supports a finalizer callback only for "wrapped" native objects.)

I'm proposing a new unified API for NAPI value references, something like this:
```C
// Set initial_refcount to 0 for a weak reference, >0 for a strong reference.
// The finalizer callback is optional.
napi_status napi_create_reference(napi_env e, napi_value value, int initial_refcount, napi_finalize cb, napi_ref* result);

// Adds a reference, optionally returning the resulting refcount. After this call the reference
// will be a strong reference because its refcount is >0, and the object is effectively "pinned".
// Calling this when the refcount is 0 and the object is unavailable results in an error.
napi_status napi_addref(napi_env e, napi_ref ref, int* result);

// Removes a reference, optionally returning the resulting refcount. If the result refcount is 0
// the reference is now weak and the object may be GC'd at any time if there are no other references.
// Calling this when the refcount is already 0 results in an error.
napi_status napi_release(napi_env e, napi_ref ref, int* result);

// Attempts to get a referenced value. If the reference is weak, the value might no longer
// be available, in that case the call is still successful but the result is NULL.
napi_status napi_get_reference(napi_env e, napi_ref ref, napi_value* result);
```

[...]

PR #99, opened 2017-02-11
Reference and external improvements (V8)

jasongin:
- Add new "external" value type and APIs
 - Replace persistent and weak-ref types and APIs with unified counted-reference `napi_ref` type and APIs
 - Update wrap/unwrap APIs to use reference type

Reviewers: I recommend looking at `node_jsvmapi.h` first to get an overview of the new/changed APIs.

This is mostly the design I proposed at https://github.com/nodejs/abi-stable-node/issues/93, with minor corrections for issues discovered during implementation.

I will update the C++ wrapper's `ObjectWrap` class to uses the updated `napi_wrap` API, now with the needed ref/unref capability via the new `napi_ref` type.

### DIFF
--- a/src/node_jsvmapi.cc
+++ b/src/node_jsvmapi.cc
@@ -153,6 +153,12 @@ namespace v8impl {
     }
 
     ~Reference() {
+      // The V8 Persistent class currently does not reset in its destructor:
+      // see NonCopyablePersistentTraits::kResetInDestructor = false.
+      // (Comments there claim that might change in the future.)
+      // To avoid memory leaks, it is better to reset at this time, however
+      // care must be taken to avoid attempting this after the Isolate has
+      // shut down, for example via a static (atexit) destructor.
       _persistent.Reset();
     }
 
@@ -187,11 +193,14 @@ namespace v8impl {
       Reference* reference = data.GetParameter();
       reference->_persistent.Reset();
 
+      // Check before calling the finalize callback, because the callback might delete it.
+      bool deleteSelf = reference->_deleteSelf;
+
       if (reference->_finalizeCallback != nullptr) {
         reference->_finalizeCallback(reference->_finalizeData);
       }
 
-      if (reference->_deleteSelf) {
+      if (deleteSelf) {
         delete reference;
       }
     }
@@ -1132,8 +1141,8 @@ napi_status napi_create_type_error(napi_env e, napi_value msg, napi_value* resul
   CHECK_ARG(result);
 
   *result = v8impl::JsValueFromV8LocalValue(
-      v8::Exception::TypeError(
-          v8impl::V8LocalValueFromJsValue(msg).As<v8::String>()));
+    v8::Exception::TypeError(
+      v8impl::V8LocalValueFromJsValue(msg).As<v8::String>()));
 
   return GET_RETURN_STATUS();
 }
--- a/src/node_jsvmapi.cc
+++ b/src/node_jsvmapi.cc
@@ -1607,17 +1607,29 @@ napi_status napi_wrap(napi_env e,
                       napi_ref* result) {
   NAPI_PREAMBLE(e);
   CHECK_ARG(jsObject);
-  CHECK_ARG(result);
 
   v8::Isolate *isolate = v8impl::V8IsolateFromJsEnv(e);
-
   v8::Local<v8::Object> obj = v8impl::V8LocalValueFromJsValue(jsObject).As<v8::Object>();
-  assert(obj->InternalFieldCount() > 0);
+
+  // Only objects that were created from a NAPI constructor's prototype
+  // via napi_define_class() can be (un)wrapped.
+  RETURN_STATUS_IF_FALSE(obj->InternalFieldCount() > 0, napi_invalid_arg);
+
   obj->SetAlignedPointerInInternalField(0, nativeObj);
 
-  v8impl::Reference* reference = new v8impl::Reference(
-    isolate, obj, 0, false, finalize_cb, nativeObj);
-  *result = reinterpret_cast<napi_ref>(reference);
+  if (finalize_cb != nullptr) {
+    // Create a separate self-deleting reference for the finalizer callback.
+    // This ensures the finalizer is not dependent on the lifetime of the returned reference.
+    new v8impl::Reference(isolate, obj, 0, true, finalize_cb, nativeObj);
+  }
+
+  if (result != nullptr) {
+    // If a reference result was requested, create one that is separate from the reference
+    // that holds the finalizer callback. The returned reference should be deleted via
+    // napi_delete_reference() when it is no longer needed.
+    v8impl::Reference* reference = new v8impl::Reference(isolate, obj, 0, false, nullptr, nullptr);
+    *result = reinterpret_cast<napi_ref>(reference);
+  }
 
   return GET_RETURN_STATUS();
 }
@@ -1629,7 +1641,10 @@ napi_status napi_unwrap(napi_env e, napi_value jsObject, void** result) {
 
   v8::Local<v8::Object> obj = v8impl::V8LocalValueFromJsValue(jsObject).As<v8::Object>();
 
-  assert(obj->InternalFieldCount() > 0);
+  // Only objects that were created from a NAPI constructor's prototype
+  // via napi_define_class() can be (un)wrapped.
+  RETURN_STATUS_IF_FALSE(obj->InternalFieldCount() > 0, napi_invalid_arg);
+
   *result = obj->GetAlignedPointerFromInternalField(0);
 
   return GET_RETURN_STATUS();
@@ -1656,9 +1671,13 @@ napi_status napi_create_external(napi_env e,
 
 napi_status napi_get_value_external(napi_env e, napi_value v, void** result) {
   NAPI_PREAMBLE(e);
+  CHECK_ARG(v);
   CHECK_ARG(result);
 
   v8::Isolate *isolate = v8impl::V8IsolateFromJsEnv(e);
+
+  v8::Local<v8::Value> value = v8impl::V8LocalValueFromJsValue(v);
+  RETURN_STATUS_IF_FALSE(value->IsExternal(), napi_invalid_arg);
 
   v8::Local<v8::External> externalValue = v8impl::V8LocalValueFromJsValue(v).As<v8::External>();
   *result = externalValue->Value();
@@ -1666,9 +1685,7 @@ napi_status napi_get_value_external(napi_env e, napi_value v, void** result) {
   return GET_RETURN_STATUS();
 }
 
-
 // Set initial_refcount to 0 for a weak reference, >0 for a strong reference.
-// The finalizer callback and callback data parameters optional.
 napi_status napi_create_reference(napi_env e,
                                   napi_value value,
                                   int initial_refcount,
--- a/src/node_jsvmapi.cc
+++ b/src/node_jsvmapi.cc
@@ -131,28 +131,78 @@ namespace v8impl {
     return u.f;
   }
 
-  static napi_persistent JsPersistentFromV8PersistentValue(
-                                             v8::Persistent<v8::Value> *per) {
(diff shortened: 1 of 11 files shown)


src/node_jsvmapi.cc
```diff
@@ -1569,122 +1600,153 @@ napi_status napi_coerce_to_string(napi_env e, napi_value v, napi_value* result)
...
+  CHECK_ARG(jsObject);
   CHECK_ARG(result);
 
-  v8::Isolate *isolate = v8impl::V8IsolateFromJsEnv(e);
-  v8::Local<v8::Context> context = isolate->GetCurrentContext();
-  v8::Local<v8::Object> obj;
-  CHECK_TO_OBJECT(context, obj, jsObject);
+  v8::Local<v8::Object> obj = v8impl::V8LocalValueFromJsValue(jsObject).As<v8::Object>();
+
+  assert(obj->InternalFieldCount() > 0);
```
boingoing: Should we catch this case as an error instead of assert? I would vote to keep as an assert. Error code is more friendly, though - if friendliness is a goal.

src/node_jsvmapi.cc
```diff
@@ -1569,122 +1600,153 @@ napi_status napi_coerce_to_string(napi_env e, napi_value v, napi_value* result)
...
+  v8::Local<v8::External> externalValue = v8impl::V8LocalValueFromJsValue(v).As<v8::External>();
+  *result = externalValue->Value();
 
   return GET_RETURN_STATUS();
 }
 
-napi_status napi_get_persistent_value(napi_env e, napi_persistent p, napi_value* result) {
+
+// Set initial_refcount to 0 for a weak reference, >0 for a strong reference.
+// The finalizer callback and callback data parameters optional.
```
boingoing: Comment seems to be stale? Don't see callback parameters.

src/node_jsvmapi.cc
```diff
@@ -1569,122 +1600,153 @@ napi_status napi_coerce_to_string(napi_env e, napi_value v, napi_value* result)
...
-                         v8::WeakCallbackType::kParameter);
-  // need to mark independent?
-  *result = v8impl::JsWeakRefFromV8PersistentValue(thePersistent);
   return GET_RETURN_STATUS();
 }
 
-napi_status napi_get_weakref_value(napi_env e, napi_weakref w, napi_value* result) {
+// Increments the reference count, optionally returning the resulting count. After this call the
+// reference will be a strong reference because its refcount is >0, and the referenced object is
+// effectively "pinned". Calling this when the refcount is 0 and the object is unavailable
```
boingoing, line 1720: Is this true about an error when the refcount is 0 and the object is unavailable? Is this an exception thrown by ClearWeak?

src/node_jsvmapi.cc
```diff
@@ -1569,122 +1600,153 @@ napi_status napi_coerce_to_string(napi_env e, napi_value v, napi_value* result)
...
   }
-  *result = v8impl::JsValueFromV8LocalValue(v);
+
   return GET_RETURN_STATUS();
 }
 
-napi_status napi_release_weakref(napi_env e, napi_weakref w) {
+// Attempts to get a referenced value. If the reference is weak, the value might no longer be
+// available, in that case the call is still successful but the result is NULL.
+napi_status napi_get_reference_value(napi_env e, napi_ref ref, napi_value* result) {
```
boingoing: I've wondered about this in the past; do you think it would be valuable to also have a bool out parameter to indicate success instead of returning nullptr for the result? Could it be the case that nullptr result is a valid (and indistinguishable from failure) value?

src/node_jsvmapi.cc
```diff
@@ -1569,122 +1600,153 @@ napi_status napi_coerce_to_string(napi_env e, napi_value v, napi_value* result)
...
+  obj->SetAlignedPointerInInternalField(0, nativeObj);
 
-  if (handle)
-  {
-    return napi_create_weakref(
-      e,
-      v8impl::JsValueFromV8LocalValue(wrap->handle()),
-      handle);
-  }
+  v8impl::Reference* reference = new v8impl::Reference(
```
boingoing: Since we're handing this reference out, user code will have to napi_delete_reference it. If caller doesn't need a handle, we could mark this one to delete itself like the Reference created for external objects. Effectively make the result optional again. Code in leveldown does actually pass nullptr for the handle, though this is probably easy to adapt.

Also, calling finalize_cb is a best effort, right? If we return this ref and calling code immediately deletes it, the persistent will be reset and callback will never be called, as I understand this. If the callback is called, then we'll be working on free'd memory which seems worse. Should we make an effort to call the callback in ~Reference() in this case? Or does persistent reset consistently call the finalizer?

boingoing (commented):
Overall looks great. Looking forward to updated ObjectWrap

src/node_jsvmapi.cc
jasongin: I'd presume the following call to `GetAlignedPointerFromInternalField(0)` should fail if the internal field count is 0. But fail how? Would it set a JavaScript exception? If so, `GET_RETURN_STATUS()` will detect that. I still don't have a good understanding of V8's error behavior, aside from maybes. Like, what if the value passed in is a primitive, or is undefined? (I remove the coercion to object because if it wasn't an object to start with then coercing to an object is not going to help with unwrapping anything.)

src/node_jsvmapi.cc
jasongin: Right, I'll fix that.

Originally I was intending for references to allow a finalize callback (in addition to externals), but then I couldn't see any way to implement it with JSRT. The `JsCreateWeakReference()` API doesn't take a callback like `JsCreateExternalObject()` does.

src/node_jsvmapi.cc
jasongin: I haven't tested yet, but I'd assume so: it shouldn't be possible to make a persistent reference to a non-existent value.

I do want to create some better test cases for this stuff.

src/node_jsvmapi.cc
jasongin: There's no situation that's indistinguishable: even if the referenced value is a JavaScript null or undefined value, the returned `napi_value` is still non-null. (It should be the same as the value returned from `napi_get_null()` or `napi_get_undfined()`.

src/node_jsvmapi.cc
jasongin: I added `CHECK_ARG(result)` at the top of this function, thinking it doesn't make sense to ever wrap an object without getting a reference to it. But maybe it should be optional, if leveldown doesn't need the handle. And then, yes I'd agree with having the weakref self-delete. (Or if there is no finalize callback provided, then there's no need to create a weakref at all.)

Yes, if the caller deletes the reference before the wrapper is GC'd then the finalizer will never get invoked. Perhaps this function should *never* return the reference, and just keep it internal and self-deleting only for the finalizer. And if the caller needs a reference, they can create one separately.

src/node_jsvmapi.cc
jasongin: This made me wonder how `napi_wrap()` is currently implemented in JSRT, when it doesn't have weakref finalize callbacks. The answer is it uses the shim (via `node::ObjectWrap`). I think we probably don't want that? But then how will we implement finalize callbacks for wrapped objects. Do we need a new JSRT API?

jasongin:
> Looking forward to updated ObjectWrap

[Already updated in the C++ wrapper PR](https://github.com/nodejs/abi-stable-node/pull/88#issuecomment-279198976)

jasongin:
They put the "Close and comment" button too close to the "Comment" button!

src/node_jsvmapi.cc
boingoing: > Perhaps this function should never return the reference, and just keep it internal and self-deleting only for the finalizer. And if the caller needs a reference, they can create one separately.

That seems good to me. Create an internal reference if there's a finalizer, otherwise make the caller explicitly request their own weak handles. Perhaps it was initially done via that optional weak handle to avoid a second function call? Or avoid having to build a second reference, though they don't seem particularly heavy-weight...

src/node_jsvmapi.cc
boingoing: I think that call to get the internal field will fail via FatalException, IIRC. Not sure if that will generate a Javascript exception or not but I think it might kill the process.

src/node_jsvmapi.cc
boingoing: JsCreateWeakReference is itself a new JSRT API, if I am remembering correctly. It hasn't flowed-back into JSRT-proper so we are free to work on it. It sounds like we should just add support for a finalizer callback. I can get behind the need for this.

src/node_jsvmapi.cc
boingoing: Fair enough and good idea, me too. :)

src/node_jsvmapi.cc
boingoing: Yeah, but napi_value is just a void*, right? An engine could use nullptr for Javascript null and just catch all the places where a conversion might happen. Not saying this is likely, or anything, just trying to propose a case. The engine would have to instead use some sentinel value when it wanted to actually use nullptr, I suppose.

src/node_jsvmapi.cc
jasongin: Then we better return an error rather than asserting. I think we should try to ensure it is not possible for *JavaScript code* to crash the engine, assuming no bugs in a native module. To hit this error condition, JavaScript code could use some different (non-wrapper) object as the `this` argument when invoking a wrapper method.

src/node_jsvmapi.cc
jasongin: Hmm. I had always assumed a valid `napi_value` would never be nullptr. I think there are a couple places in the C++ wrapper code that rely on that. While I don't like it, I think your point is valid, unless we want to explicitly take a stand against nullptr values.

Regardless of that, why would anyone ever need to create a reference to the value null? It's not like it can ever change or go away.

src/node_jsvmapi.cc
jasongin: The problem is the ObjectWrap C++ wrapper will always need a reference, because it needs be able to add refcounts (to make the object persistent during async callbacks). So that would be inefficient. Maybe we just need to document that the returned reference should not be deleted until the finalize callback?

jasongin: I'm going to compromise here and keep the reference out parameter of `napi_wrap`, but make it optional, and *separate* from the self-deleting reference that manages the finalizer callback. That way the finalizer will not be dependent on the lifetime of the received reference, but object wrapping won't require an extra NAPI call to get a reference.

boingoing (approved):
Looks good 👍

mhdawson:
So I'm convinced its cleaner this way than re-using node's object wrap.  I also like that we still have wrap/unwrap (both to avoid more calls) but also so that when somebody's porting its more obvious how they can port wrap and unwrap.

One remaining question is if we are actually breaking the Chakra implementation by not using node's object wrap.  Currently its shimmed and all works ok.  Sounds like with this implementation we need changes in Chakra before it will work completely ? Or did I just misread the comments?

mhdawson (commented):
Generally looks good, just the remaining comment about where we are introducing an issue with Chakra or not ?

boingoing:
@mhdawson 
>are actually breaking the Chakra implementation by not using node's object wrap. Currently its shimmed and all works ok. Sounds like with this implementation we need changes in Chakra before it will work completely ?

Yes, weak refs do not support a finalizer callback in ChakraCore yet. We should be able to add support since the Jsrt weak reference APIs are part of this branch. We have to update the Jsrt branch anyway to support these NAPI API changes so seems likely to add support at the same time.

jasongin:
> Sounds like with this implementation we need changes in Chakra before it will work completely ?

The `napi_wrap` implementation for ChakraCore can continue using the shim (via `node::ObjectWrap`) as it already does, until we add the necessary JSRT API support. It will continue to work that way, but we'll want to eliminate reliance on the shim eventually.

PR #101, opened 2017-02-11
Property & class definition improvements (ChakraCore)

jasongin:
- Rename napi_create_constructor to napi_define_class and include support for defining static properties on the class at the same time
 - Rename napi_define_property to napi_define_properties to allow for defining multiple properties at the same time
 - Update tests for the changes

Most of the files are copied from the [same change in the V8 branch](https://github.com/nodejs/abi-stable-node/pull/97), except for the JSRT implementation which is new.

### DIFF
--- a/src/node_jsrtapi.cc
+++ b/src/node_jsrtapi.cc
@@ -294,8 +294,13 @@ napi_status napi_create_function(napi_env e, napi_callback cb, void* data, napi_
   return napi_ok;
 }
 
-napi_status napi_create_constructor(napi_env e, const char* utf8name, napi_callback cb,
-  void* data, int property_count, napi_property_descriptor* properties, napi_value* result) {
+napi_status napi_define_class(napi_env e,
+                              const char* utf8name,
+                              napi_callback cb,
+                              void* data,
+                              int property_count,
+                              const napi_property_descriptor* properties,
+                              napi_value* result) {
   CHECK_ARG(result);
 
   napi_value namestring;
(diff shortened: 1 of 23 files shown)


jasongin:
Rebased. Can someone review this please?

gabrielschulhof:
Seems `ChakraCore.h` needs to be included somewhere: https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-commit-linux-abi/nodes=ubuntu1404-64/7/console

jasongin:
> ../src/node_jsrtapi.cc:26:24: fatal error: ChakraCore.h: No such file or directory
 #include "ChakraCore.h"

That failure is unrelated to this change, because this change certainly did not introduce the dependency on "ChakraCore.h", or change any build configuration.

src/node_jsrtapi.cc
```diff
@@ -315,8 +320,42 @@ napi_status napi_create_constructor(napi_env e, const char* utf8name, napi_callb
...
+    if ((properties[i].attributes & napi_static_property) != 0) {
+      staticPropertyCount++;
+    }
+    else {
+      instancePropertyCount++;
+    }
+  }
+
+  std::vector<napi_property_descriptor> descriptors;
+  descriptors.reserve(instancePropertyCount > staticPropertyCount ?
```
digitalinfinity: Nit: since we have an stl dependency anyway, can we use max here - makes the code just a tad bit more readable

src/node_jsrtapi.cc
```diff
@@ -364,69 +403,78 @@ napi_status napi_set_property(napi_env e, napi_value o,
...
-    napi_property_descriptor* p) {
-  napi_value descriptor;
-  CHECK_NAPI(napi_create_object(e, &descriptor));
-
+napi_status napi_define_properties(napi_env e,
+                                   napi_value o,
+                                   int property_count,
+                                   const napi_property_descriptor* properties) {
   napi_propertyname configurableProperty;
   CHECK_NAPI(napi_property_name(e, "configurable", &configurableProperty));
```
digitalinfinity: Not related to your PR but I wonder if this should be napi_get_property_name, to be consistent with other NAPI APIs.

Also, more generally, I wonder what our philosophy should be with respect to NAPI implementations- this method for example generally calls other NAPI APIs- is there a reason for that vs. straight up calling JSRT APIs? In the end, I'm not sure it matters for the trivial NAPI wrappers since the optimizer should be able to inline those calls, but I was more curious if this decision was an intentional one or not.

digitalinfinity:
:shipit:

digitalinfinity (approved):

src/node_jsrtapi.cc
jasongin: > I wonder if this should be napi_get_property_name, to be consistent with other NAPI APIs.

Actually @boingoing and I have discussed eliminating the `napi_propertyname` type, because it seems to be a JSRT concept that leaked into the NAPI public API. V8 allows using any value type as a property key, and apparently leveldown makes use of that.

> this method calls other NAPI APIs- is there a reason for that vs. straight up calling JSRT APIs

Just for convenience & maintainability. Why not?

digitalinfinity: I'd be in favor of eliminating napi_propertyname too. @gabrielschulhof @mhdawson @sampsongao thoughts?

Convenience I can see- what is the maintainability argument? I don't have strong feelings one way or the other but it does seem like an additional layer of abstraction that could hurt perf in some cases- and it makes the code slightly weirder to read, seeing a mix of JSRT and NAPI calls in a single method implementation- I can see the calls to NAPI if the equivalent JSRT methods don't exist, but in the event that they do, it triggers the OCD part of me 😄

jasongin: Since the calls to NAPI APIs here are all basically trivial wrappers around the JSRT APIs, I guess it's not saving much code if any. So I'd agree it would be better to rewrite this function using direct calls to JSRT.

jasongin:
> ../src/node_jsrtapi.cc:26:24: fatal error: ChakraCore.h: No such file or directory
 #include "ChakraCore.h"

The CI build is mssing the `--engine=chakracore` parameter to `./configure`. We haven't yet merged [the node-chakracore change that makes chakracore the default engine](https://github.com/nodejs/node-chakracore/commit/539ff9476b58285246353fe1fcd5858034606953).

src/node_jsrtapi.cc
gabrielschulhof: I'm also in favour of eliminating `napi_propertyname`.

jasongin:
Updated to address @digitalinfinity's comments.

I'm going to go ahead and merge this since he already signed off.

PR #104, opened 2017-02-14
build: size_t needs stdlib.h on linux

gabrielschulhof:


src/node_jsrtapi.cc
```diff
@@ -17,6 +17,7 @@
 *
 ******************************************************************************/
 
+#include <stdlib.h>
```
bnoordhuis: stddef.h should be enough.

gabrielschulhof: @bnoordhuis Thanks!

mhdawson (approved):
LGTM

PR #106, opened 2017-02-15
napi: function may be called with any value as recv

gabrielschulhof:
We should not assert that the context for a function call is an object.


mhdawson (approved):
Looking at the definition for Call in v8 does seem to take all sorts of types and your test would confirm that  as well.  LGTM.

PR #107, opened 2017-02-15
napi: synchronize function call context test

gabrielschulhof:
Make the test perform the same operations it does on V8.


mhdawson (approved):
LGTM, tests should be the same across repos.

Issue #108, opened 2017-02-20
We need a version of `napi_buffer_new()` where a deleter is given


gabrielschulhof:
`node::Buffer::New()` has a version where it takes a `char *` along with the size. Node assumes ownership of the memory at the `char *` and frees it using `free()` when the buffer is garbage collected. It also provides a version of `New()` where a deleter may be specified in case `free()` is not the right thing to do upon GC.

To avoid excessive copying, addons should be able to set the deleter to a no-op. To that end we need a version of `napi_buffer_new()` that accepts a deleter. I propose
```C
napi_status napi_buffer_new_with_deleter(napi_env e,
                                         char* data,
                                         size_t size,
                                         void (*deleter)(void*),
                                         napi_value* result);
```

gabrielschulhof:
Forgot the context ...
```C++
napi_status napi_buffer_new_with_deleter(napi_env e,
                                         char* data,
                                         size_t size,
                                         void (*deleter)(char*, void*),
                                         void* deleter_context,
                                         napi_value* result);
```

jasongin:
I agree, it seems like a good API to add. There are a few details to consider.

* The buffer APIs are currently named inconsistently with everything else, e.g. `napi_buffer_new` instead of `napi_create_buffer`.
* It should probably use `napi_finalize` to be consistent with the deleter callbacks used by `napi_wrap` and `napi_create_external`. `napi_finalize` doesn't currently take a context, aside from the pointer that is to be deleted. Does it need a context?
* There is a similar API already for creating arraybuffers on external data `napi_create_external_arraybuffer` however it is missing a deleter callback, that it probably should have for the same reasons.

gabrielschulhof:
Well, node::Buffer deleters take a context, but I guess we needn't expose
that.

gabrielschulhof:
TBH, I'm not sure we need `napi_buffer_new()` at all. In fact, we should have
```C++
napi_status napi_create_buffer(napi_env e,
                               char* data,
                               size_t size,
                               napi_finalize finalize_cb,
                               napi_value* result);
```
and no other variation on `napi_create_buffer()`. We can document that `finalize_cb` may be null, in which case `free()` will be called on the pointer upon gc.

jasongin:
> We can document that finalize_cb may be null, in which case free() will be called on the pointer upon gc.

[...]

PR #109, opened 2017-02-20
napi: revamp buffer API

gabrielschulhof:
Fixes gh-108

### DIFF
--- a/src/node_jsvmapi.cc
+++ b/src/node_jsvmapi.cc
@@ -1951,21 +1951,33 @@ napi_status napi_get_and_clear_last_exception(napi_env e, napi_value* result) {
   return napi_ok;
 }
 
-napi_status napi_create_buffer(napi_env e,
-                               char* data,
-                               size_t size,
-                               napi_finalize finalize_cb,
+napi_status napi_create_buffer(napi_env e, size_t size, char** data,
                                napi_value* result) {
   NAPI_PREAMBLE(e);
+  CHECK_ARG(data);
   CHECK_ARG(result);
 
-  v8::MaybeLocal<v8::Object> maybe;
-  if (finalize_cb) {
-    maybe = node::Buffer::New(v8impl::V8IsolateFromJsEnv(e), data, size,
-      (node::Buffer::FreeCallback)finalize_cb, nullptr);
-  } else {
-    maybe = node::Buffer::New(v8impl::V8IsolateFromJsEnv(e), data, size);
-  }
+  auto maybe = node::Buffer::New(v8impl::V8IsolateFromJsEnv(e), size);
+
+  CHECK_MAYBE_EMPTY(maybe, napi_generic_failure);
+
+  v8::Local<v8::Object> jsBuffer = maybe.ToLocalChecked();
+
+  *result = v8impl::JsValueFromV8LocalValue(jsBuffer);
+  *data = node::Buffer::Data(jsBuffer);
+
+  return GET_RETURN_STATUS();
+}
+
+napi_status napi_create_external_buffer(napi_env e, size_t size, char* data,
+                                        napi_finalize finalize_cb,
+                                        napi_value* result) {
+  NAPI_PREAMBLE(e);
+  CHECK_ARG(data);
+  CHECK_ARG(result);
+
+  auto maybe = node::Buffer::New(v8impl::V8IsolateFromJsEnv(e), data, size,
+    (node::Buffer::FreeCallback)finalize_cb, nullptr);
 
   CHECK_MAYBE_EMPTY(maybe, napi_generic_failure);
 
@@ -1994,19 +2006,20 @@ napi_status napi_is_buffer(napi_env e, napi_value v, bool* result) {
   return GET_RETURN_STATUS();
 }
 
-napi_status napi_get_buffer_data(napi_env e, napi_value v, char** result) {
+napi_status napi_get_buffer_info(napi_env e, napi_value v, char** data,
+                                 size_t* length) {
   NAPI_PREAMBLE(e);
-  CHECK_ARG(result);
 
-  *result = node::Buffer::Data(v8impl::V8LocalValueFromJsValue(v).As<v8::Object>());
-  return GET_RETURN_STATUS();
-}
+  v8::Local<v8::Object> buffer =
+    v8impl::V8LocalValueFromJsValue(v).As<v8::Object>();
 
-napi_status napi_get_buffer_length(napi_env e, napi_value v, size_t* result) {
-  NAPI_PREAMBLE(e);
-  CHECK_ARG(result);
+  if (data) {
+    *data = node::Buffer::Data(buffer);
+  }
+  if (length) {
+    *length = node::Buffer::Length(buffer);
+  }
 
-  *result = node::Buffer::Length(v8impl::V8LocalValueFromJsValue(v));
   return GET_RETURN_STATUS();
 }
 
--- a/src/node_jsvmapi.cc
+++ b/src/node_jsvmapi.cc
@@ -1951,11 +1951,21 @@ napi_status napi_get_and_clear_last_exception(napi_env e, napi_value* result) {
   return napi_ok;
 }
 
-napi_status napi_buffer_new(napi_env e, char* data, size_t size, napi_value* result) {
+napi_status napi_create_buffer(napi_env e,
+                               char* data,
+                               size_t size,
+                               napi_finalize finalize_cb,
+                               napi_value* result) {
   NAPI_PREAMBLE(e);
   CHECK_ARG(result);
 
-  auto maybe = node::Buffer::New(v8impl::V8IsolateFromJsEnv(e), data, size);
+  v8::MaybeLocal<v8::Object> maybe;
+  if (finalize_cb) {
+    maybe = node::Buffer::New(v8impl::V8IsolateFromJsEnv(e), data, size,
+      (node::Buffer::FreeCallback)finalize_cb, nullptr);
+  } else {
+    maybe = node::Buffer::New(v8impl::V8IsolateFromJsEnv(e), data, size);
+  }
 
   CHECK_MAYBE_EMPTY(maybe, napi_generic_failure);
 
@@ -1963,8 +1973,8 @@ napi_status napi_buffer_new(napi_env e, char* data, size_t size, napi_value* res
   return GET_RETURN_STATUS();
 }
 
-napi_status napi_buffer_copy(napi_env e, const char* data,
-                             size_t size, napi_value* result) {
+napi_status napi_create_buffer_copy(napi_env e, const char* data,
+                                    size_t size, napi_value* result) {
   NAPI_PREAMBLE(e);
   CHECK_ARG(result);
 
@@ -1976,7 +1986,7 @@ napi_status napi_buffer_copy(napi_env e, const char* data,
   return GET_RETURN_STATUS();
 }
 
-napi_status napi_buffer_has_instance(napi_env e, napi_value v, bool* result) {
+napi_status napi_is_buffer(napi_env e, napi_value v, bool* result) {
   NAPI_PREAMBLE(e);
   CHECK_ARG(result);
 
@@ -1984,7 +1994,7 @@ napi_status napi_buffer_has_instance(napi_env e, napi_value v, bool* result) {
   return GET_RETURN_STATUS();
 }
 
-napi_status napi_buffer_data(napi_env e, napi_value v, char** result) {
+napi_status napi_get_buffer_data(napi_env e, napi_value v, char** result) {
   NAPI_PREAMBLE(e);
   CHECK_ARG(result);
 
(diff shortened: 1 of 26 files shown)


src/node_jsvmapi.h
```diff
@@ -315,17 +315,18 @@ NODE_EXTERN napi_status napi_is_exception_pending(napi_env e, bool* result);
...
-                                        size_t size,
-                                        napi_value* result);
-NODE_EXTERN napi_status napi_buffer_copy(napi_env e,
-                                         const char* data,
-                                         size_t size,
-                                         napi_value* result);
-NODE_EXTERN napi_status napi_buffer_has_instance(napi_env e, napi_value v, bool* result);
-NODE_EXTERN napi_status napi_buffer_data(napi_env e, napi_value v, char** result);
-NODE_EXTERN napi_status napi_buffer_length(napi_env e, napi_value v, size_t* result);
+NODE_EXTERN napi_status napi_create_buffer(napi_env e,
```
jasongin, line 318: I'd actually prefer 2 separate APIs:
 * `napi_create_buffer` allocates and frees the buffer for you.
 * `napi_create_external_buffer` takes an external data pointer and finalize callback.

That would be consistent with the `napi_*_arraybuffer` APIs. Except `napi_create_external_arraybuffer` is missing a finalize callback. That was an oversight in my implementation. [Added here.](https://github.com/nodejs/abi-stable-node/pull/110)

JS `ArrayBuffer` and `node::Buffer` are almost the same thing anyway. As far as I know, the only reason node has its own buffer type is because the JS type didn't exist when Node got started.

src/node_jsvmapi.h
```diff
@@ -315,17 +315,18 @@ NODE_EXTERN napi_status napi_is_exception_pending(napi_env e, bool* result);
...
+                                           size_t size,
+                                           napi_finalize finalize_cb,
+                                           napi_value* result);
+NODE_EXTERN napi_status napi_create_buffer_copy(napi_env e,
+                                                const char* data,
+                                                size_t size,
+                                                napi_value* result);
+NODE_EXTERN napi_status napi_is_buffer(napi_env e, napi_value v, bool* result);
+NODE_EXTERN napi_status napi_get_buffer_data(napi_env e, napi_value v, char** result);
+NODE_EXTERN napi_status napi_get_buffer_length(napi_env e, napi_value v, size_t* result);
```
jasongin: Should the 2 functions for getting data and length be combined into a single `napi_get_buffer_info`, like `napi_get_arraybuffer_info` ?

src/node_jsvmapi.h
mhdawson: I'm + 1 on having the two methods as they seem different enough.

src/node_jsvmapi.h
mhdawson: Less rounds trips would be good.

kkoopa:
You should also allow passing the `hint` argument of `node::Buffer` and `node::FreeCallback`.

jasongin:
> You should also allow passing the hint argument of node::Buffer and node::FreeCallback.

Probably. That would require updating `napi_finalize` to include a hint parameter, and updating all APIs that use it to also take a hint.

test/addons-abi/test_buffer/test_buffer.cc
```diff
@@ -0,0 +1,141 @@
...
+}
+
+NAPI_METHOD(bufferHasInstance) {
+  int argc;
+  NAPI_CALL(env, napi_get_cb_args_length(env, info, &argc));
+  JS_ASSERT(env, argc == 1, "Wrong number of arguments");
+  napi_value theBuffer;
+  NAPI_CALL(env, napi_get_cb_args(env, info, &theBuffer, 1));
+  bool hasInstance;
+  NAPI_CALL(env, napi_is_buffer(env, theBuffer, &hasInstance));
```
jasongin: Assert `hasInstance == true`.

jasongin: Can you also add a call to `napi_get_type_of_value()` on the buffer? The expected result should be `napi_object`, except I think it might fail with JSRT. Due to use of the shim there the buffer `napi_value` is not a `JsValueRef`, so `JsGetValueType()` will fail, or maybe crash.

test/addons-abi/test_buffer/test_buffer.cc
```diff
@@ -0,0 +1,141 @@
+#include <string.h>
+#include <string>
+#include <node_api_helpers.h>
+
+#define JS_ASSERT(env, assertion, message) \
```
jasongin, line 5: This macro looks useful! Should it go in a header shared by all the tests? Maybe also the next macro.

test/addons-abi/test_buffer/test_buffer.cc
```diff
@@ -0,0 +1,141 @@
...
+  if (!(assertion)) { \
+    napi_throw_error((env), (std::string("assertion (" #assertion ") failed: ") + message).c_str()); \
```
jasongin, line 7: Is it possible for the compiler to concatenate the message string also?

test/addons-abi/test_buffer/test_buffer.cc
```diff
@@ -0,0 +1,141 @@
...
+  NAPI_CALL(env, napi_create_function(env, bufferData, nullptr, &theValue));
+  NAPI_CALL(env, napi_set_property(env, exports, propName, theValue));
+
+  NAPI_CALL(env, napi_property_name(env, "staticBuffer", &propName));
+  NAPI_CALL(env, napi_create_function(env, staticBuffer, nullptr, &theValue));
+  NAPI_CALL(env, napi_set_property(env, exports, propName, theValue));
+
+  NAPI_CALL(env, napi_property_name(env, "bufferLength", &propName));
+  NAPI_CALL(env, napi_create_function(env, bufferLength, nullptr, &theValue));
+  NAPI_CALL(env, napi_set_property(env, exports, propName, theValue));
```
jasongin: All these properties could be set up more cleanly with a single call to `napi_define_properties()`. See the other tests' `Init()` functions.

test/addons-abi/1_hello_world/test.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
-require('../../common');
+var common = require('../../common');
 var assert = require('assert');
-var addon = require('./build/Release/binding');
+var addon = require(`./build/${common.buildType}/binding`);
```
jasongin, line 4: Thank you! I had been wondering if there was a good way to do this.

test/addons-abi/test_buffer/test_buffer.cc
gabrielschulhof: I'm not certain about how generically useful it is, because it can only be called from a binding, which returns void.

I have a pile of macros for common operations [here](https://github.com/gabrielschulhof/iotivity-node/blob/f09b6ae0c51d8b653bd32f51b72ac5d3b4c3ec73/src/common.h). If you think any of them could be generically useful, I will add them.

test/addons-abi/test_buffer/test_buffer.cc
gabrielschulhof: I don't understand. It does get concatenated with the `std::string` that precedes it.

gabrielschulhof:
@kkoopa I've filed https://github.com/nodejs/abi-stable-node/issues/111 regarding your suggestion. I suggest we leave the `napi_finalize` callbacks in this PR without a context parameter and make the fix for #111 a separate PR which addresses all `napi_finalize` callback use.

test/addons-abi/test_buffer/test_buffer.cc
gabrielschulhof: I'll file an issue to pass over the tests and use macros. I've distilled [a bunch of macros](https://github.com/gabrielschulhof/iotivity-node/blob/f09b6ae0c51d8b653bd32f51b72ac5d3b4c3ec73/src/common.h) that come in handy with this API, though those are more generic than we might need for the tests, because in the tests we're not likely to have helper functions.

test/addons-abi/test_buffer/test_buffer.cc
```diff
@@ -60,45 +70,35 @@ NAPI_METHOD(bufferHasInstance) {
...
+	&returnValue));
   NAPI_CALL(env, napi_set_return_value(env, info, returnValue));
 }
 
 NAPI_METHOD(staticBuffer) {
   napi_value theBuffer;
-  NAPI_CALL(env, napi_create_buffer(env, (char *)theText,
-    sizeof(theText), noopDeleter, &theBuffer));
+  NAPI_CALL(env, napi_create_external_buffer(env, sizeof(theText),
+    (char *)theText, noopDeleter, &theBuffer));
```
jasongin: This can use `nullptr` now instead of `noopDeleter`.

src/node_jsvmapi.h
```diff
@@ -316,17 +316,20 @@ NODE_EXTERN napi_status napi_get_and_clear_last_exception(napi_env e, napi_value
...
                                            size_t size,
-                                           napi_finalize finalize_cb,
+                                           char** data,
                                            napi_value* result);
+NODE_EXTERN napi_status napi_create_external_buffer(napi_env e,
+                                                    size_t size,
+													char* data,
+													napi_finalize finalize_cb,
+													napi_value* result);
 NODE_EXTERN napi_status napi_create_buffer_copy(napi_env e,
```
jasongin: Replace tabs with spaces.

src/node_jsvmapi.cc
```diff
@@ -1951,21 +1951,33 @@ napi_status napi_get_and_clear_last_exception(napi_env e, napi_value* result) {
...
 }
 
-napi_status napi_create_buffer(napi_env e,
-                               char* data,
-                               size_t size,
-                               napi_finalize finalize_cb,
+napi_status napi_create_buffer(napi_env e, size_t size, char** data,
                                napi_value* result) {
   NAPI_PREAMBLE(e);
+  CHECK_ARG(data);
```
jasongin: The `data` parameter can be optional.

test/addons-abi/test_buffer/test_buffer.cc
jasongin: This is not done in the PR update. Did you overlook this comment?
I think it's likely to catch a bug.

test/addons-abi/test_buffer/test_buffer.cc
```diff
@@ -0,0 +1,124 @@
...
+  napi_value theValue;
+
+  NAPI_CALL(env, napi_property_name(env, "theText", &propName));
+  NAPI_CALL(env, napi_create_string_utf8(env, theText, sizeof(theText), &theValue));
+  NAPI_CALL(env, napi_set_property(env, exports, propName, theValue));
+
+  napi_property_descriptor methods[] = {
+    { "newBuffer", newBuffer },
+	{ "newExternalBuffer", newExternalBuffer },
+	{ "getDeleterCallCount", getDeleterCallCount },
```
jasongin: Replace tabs with spaces in this file.

test/addons-abi/test_buffer/test_buffer.cc
jasongin: If the message is a string literal, then it can be concatenated at compile time, just like `#assertion`.

But I suppose requiring a string literal might not be ideal. And it's not like we care about performance here. So, feel free to disregard this suggestion.

gabrielschulhof: Yeah, and this way it can accommodate both a literal and a std::string.

test/addons-abi/test_buffer/test_buffer.cc
gabrielschulhof: Sorry! I forgot.

src/node_jsvmapi.h
```diff
@@ -315,17 +315,21 @@ NODE_EXTERN napi_status napi_is_exception_pending(napi_env e, bool* result);
...
-NODE_EXTERN napi_status napi_buffer_has_instance(napi_env e, napi_value v, bool* result);
-NODE_EXTERN napi_status napi_buffer_data(napi_env e, napi_value v, char** result);
-NODE_EXTERN napi_status napi_buffer_length(napi_env e, napi_value v, size_t* result);
+NODE_EXTERN napi_status napi_create_buffer(napi_env e,
+                                           size_t size,
+                                           char** data,
+                                           napi_value* result);
+NODE_EXTERN napi_status napi_create_external_buffer(napi_env e,
+                                                    size_t size,
+													char* data,
```
mhdawson, line 324: This is new so we should line it up as part of this PR right ?

gabrielschulhof: What do you mean by "line it up"?

jasongin: He's referring to the messy indentation due to tabs. The last commit removed tabs below (where I commented) but not here.

jasongin:
LGTM
Just fix the tabs. :)

PR #110, opened 2017-02-21
Add finalize cb to external arraybuffer (V8)

jasongin:
We should support a finalize callback wherever an external data pointer is provided. This was an oversight in my TypedArray implementation.

I've also added a test case for the `napi_create_external_arraybuffer` API, though it doesn't use the finalize callback because it would be difficult to force the finalizer to be invoked during the test.


boingoing (approved):
LGTM

PR #113, opened 2017-02-22
Reference and external improvements (ChakraCore)

jasongin:
- Add APIs for external values with data pointer & finalizer callback
 - Replace persistent and weak-ref types and APIs with unified counted-reference type and APIs
 - Update wrap/unwrap APIs to use reference type and avoid using the shim
 - Add finalize callback for external arraybuffer
 - Fix handling of data context for function callbacks

Note the changes in `node_jsvmapi.h`, `node_api_helpers.h`, and test files are just copied from [the equivalent change in the V8 branch](https://github.com/nodejs/abi-stable-node/pull/99).

### DIFF
--- a/src/node_jsrtapi.cc
+++ b/src/node_jsrtapi.cc
@@ -180,7 +180,7 @@ namespace jsrtimpl {
       cbInfo.isConstructCall = isConstructCall;
 
       if (isConstructCall) {
-        // For constructor callbacks, replace the the 'this' arg with a new external object,
+        // For constructor callbacks, replace the 'this' arg with a new external object,
         // to support wrapping a native object in the external object.
         JsValueRef externalThis;
         if (JsNoError == JsCreateExternalObject(
@@ -319,6 +319,7 @@ napi_status napi_create_function(napi_env e, napi_callback cb, void* data, napi_
   CHECK_ARG(result);
 
   jsrtimpl::ExternalCallback* externalCallback = new jsrtimpl::ExternalCallback(cb, data);
+  if (externalCallback == nullptr) return napi_set_last_error(napi_generic_failure);
 
   JsValueRef function;
   CHECK_JSRT(JsCreateFunction(jsrtimpl::ExternalCallback::Callback, externalCallback, &function));
@@ -343,6 +344,7 @@ napi_status napi_define_class(napi_env e,
   CHECK_NAPI(napi_create_string_utf8(e, utf8name, -1, &namestring));
 
   jsrtimpl::ExternalCallback* externalCallback = new jsrtimpl::ExternalCallback(cb, data);
+  if (externalCallback == nullptr) return napi_set_last_error(napi_generic_failure);
 
   JsValueRef constructor;
   CHECK_JSRT(JsCreateNamedFunction(
@@ -1051,6 +1053,8 @@ napi_status napi_wrap(napi_env e, napi_value jsObject, void* nativeObj,
   JsValueRef value = reinterpret_cast<JsValueRef>(jsObject);
 
   jsrtimpl::ExternalData* externalData = new jsrtimpl::ExternalData(nativeObj, finalize_cb);
+  if (externalData == nullptr) return napi_set_last_error(napi_generic_failure);
+
   CHECK_JSRT(JsSetExternalData(value, externalData));
 
   if (result != nullptr) {
@@ -1068,8 +1072,11 @@ napi_status napi_create_external(napi_env e, void* data, napi_finalize finalize_
     napi_value* result) {
   CHECK_ARG(result);
 
+  jsrtimpl::ExternalData* externalData = new jsrtimpl::ExternalData(data, finalize_cb);
+  if (externalData == nullptr) return napi_set_last_error(napi_generic_failure);
+
   CHECK_JSRT(JsCreateExternalObject(
-    new jsrtimpl::ExternalData(data, finalize_cb),
+    externalData,
     jsrtimpl::ExternalData::Finalize,
     reinterpret_cast<JsValueRef*>(result)));
 
@@ -1395,12 +1402,15 @@ napi_status napi_create_external_arraybuffer(napi_env e,
                                              napi_value* result) {
   CHECK_ARG(result);
 
+  jsrtimpl::ExternalData* externalData = new jsrtimpl::ExternalData(external_data, finalize_cb);
+  if (externalData == nullptr) return napi_set_last_error(napi_generic_failure);
+
   JsValueRef arrayBuffer;
   CHECK_JSRT(JsCreateExternalArrayBuffer(
     external_data,
     static_cast<unsigned int>(byte_length),
     jsrtimpl::ExternalData::Finalize,
-    new jsrtimpl::ExternalData(external_data, finalize_cb),
+    externalData,
     &arrayBuffer));
 
   *result = reinterpret_cast<napi_value>(arrayBuffer);
--- a/src/node_jsrtapi.cc
+++ b/src/node_jsrtapi.cc
@@ -39,6 +39,7 @@ void napi_module_register(void* mod) {
 //Callback Info struct as per JSRT native function.
 struct CallbackInfo {
   bool isConstructCall;
+  napi_value thisArg;
   unsigned short argc;
   napi_value* argv;
   void* data;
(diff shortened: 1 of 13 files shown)


test/addons-abi/3_callbacks/binding.cc
```diff
@@ -41,8 +41,7 @@ void Init(napi_env env, napi_value exports, napi_value module) {
     { "RunCallback", RunCallback },
     { "RunCallbackWithRecv", RunCallbackWithRecv }
   };
-  status = napi_define_properties(
-    env, exports, sizeof(desc) / sizeof(napi_property_descriptor), desc);
+  status = napi_define_properties(env, exports, 2, desc);
```
digitalinfinity, line 44: if its always an array passed in, could we have a convenience macro for this that passes in the count- also, super-nitpicky, but generally, APIs that take in an array and count, take the array parameter first

src/node_jsrtapi.cc
```diff
@@ -126,51 +127,100 @@ namespace v8impl {
...
+      ExternalData* externalData = reinterpret_cast<ExternalData*>(callbackState);
+      if (externalData != nullptr) {
+        if (externalData->_cb != nullptr) {
+          externalData->_cb(externalData->_data);
+        }
 
-  void* getValue() {
-    return _nativeObj;
-  }
+        delete externalData;
```
digitalinfinity, line 150: We should document that the data is always allocated using the CRT heap? otherwise we should take in an extra parameter to free the data too

src/node_jsrtapi.cc
```diff
@@ -126,51 +127,100 @@ namespace v8impl {
...
+        JsValueRef externalThis;
+        if (JsNoError == JsCreateExternalObject(
+          nullptr, jsrtimpl::ExternalData::Finalize, &externalThis)) {
+
+          // Copy the prototype from the default 'this' arg to the new 'external' this arg.
+          if (arguments[0] != nullptr) {
+            JsValueRef thisPrototype;
+            if (JsNoError == JsGetPrototype(arguments[0], &thisPrototype)) {
+              JsSetPrototype(externalThis, thisPrototype);
+            }
```
digitalinfinity, line 194: What happens ifJsGetPrototype/JsSetPrototype fails

src/node_jsrtapi.cc
```diff
@@ -126,51 +127,100 @@ namespace v8impl {
...
+      JsErrorCode error = JsNoError;
+      JsValueRef undefinedValue;
+      error = JsGetUndefinedValue(&undefinedValue);
+
+      CallbackInfo cbInfo;
+      cbInfo.thisArg = reinterpret_cast<napi_value>(arguments[0]);
+      cbInfo.isConstructCall = isConstructCall;
+
+      if (isConstructCall) {
+        // For constructor callbacks, replace the the 'this' arg with a new external object,
```
digitalinfinity: Nit- extra 'the'

src/node_jsrtapi.cc
```diff
@@ -1004,67 +1047,157 @@ napi_status napi_coerce_to_string(napi_env e, napi_value v, napi_value* result)
...
 
 napi_status napi_unwrap(napi_env e, napi_value jsObject, void** result) {
-  CHECK_ARG(result);
-  *result = ObjectWrapWrapper::Unwrap(jsObject);
-  return napi_ok;
+  return napi_get_value_external(e, jsObject, result);
 }
 
-napi_status napi_create_persistent(napi_env e, napi_value v, napi_persistent* result) {
+napi_status napi_create_external(napi_env e, void* data, napi_finalize finalize_cb,
```
digitalinfinity, line 1071: Nit- mildly prefer napi_create_with_external_data but it's also unwieldy so I could go either way

src/node_jsrtapi.cc
```diff
@@ -1004,67 +1047,157 @@ napi_status napi_coerce_to_string(napi_env e, napi_value v, napi_value* result)
 }
 
 napi_status napi_wrap(napi_env e, napi_value jsObject, void* nativeObj,
-  napi_destruct destructor, napi_weakref* handle) {
-  new ObjectWrapWrapper(jsObject, nativeObj, destructor);
+    napi_finalize finalize_cb, napi_ref* result) {
+  JsValueRef value = reinterpret_cast<JsValueRef>(jsObject);
```
digitalinfinity, line 1053: Does this jsObject have to be JsrtExternalObject? Is this something that would behave differently between v8 and chakra? Should this be modelled somehow by the type system?

src/node_jsrtapi.cc
```diff
@@ -1004,67 +1047,157 @@ napi_status napi_coerce_to_string(napi_env e, napi_value v, napi_value* result)
...
+napi_status napi_create_reference(napi_env e, napi_value v, int initial_refcount,
+    napi_ref* result) {
   CHECK_ARG(result);
+
   JsValueRef strongRef = reinterpret_cast<JsValueRef>(v);
   JsWeakRef weakRef = nullptr;
   CHECK_JSRT(JsCreateWeakReference(strongRef, &weakRef));
+
+  // Prevent the reference itself from being collected until it is explicitly deleted.
   CHECK_JSRT(JsAddRef(weakRef, nullptr));
```
digitalinfinity, line 1109: So for every reference you create, you need to create a weak reference, and then addref than and the strong reference? This seems unnecessarily expensive, no? One idea is that you addref the strong reference instead, and install a beforecollect callback to null out some location to indicate when the strong reference is collected (effectively model Chakra's weak reference handles in napi). An alternative is that you always have a reference to the weak reference handle (that is, it has a refcount of 1 always), and you only addref the strong ref- and you decrement the weak ref handle only in delete_reference?

src/node_jsvmapi_types.h
```diff
@@ -51,6 +50,7 @@ enum napi_valuetype {
   napi_symbol,
   napi_object,
   napi_function,
+  napi_external,
```
digitalinfinity, line 53: This is a weird object type since it's not quite an ES6 type- should this be something like napi_object_with_external_data to further clarify its relation to an ES6 type? Or should this be a modifier bit so that any es6 type can be created with external data (i.e napi_valuetypes have high-order modifier bits to control creation of the object- this could enable other things like creating arbitrary objects with special napi hooks etc)

test/addons-abi/7_factory_wrap/myobject.h
```diff
@@ -13,10 +13,12 @@ class MyObject {
   MyObject();
   ~MyObject();
 
-  static napi_persistent constructor;
+  static napi_ref constructor;
   static void New(napi_env env, napi_callback_info info);
   static void PlusOne(napi_env env, napi_callback_info info);
   double counter_;
+  napi_env env_;
```
digitalinfinity, line 20: Just an api ergonomics question- can we store the environment in the wrapped object so that people don't need to store a reference to the environment for use later?

src/node_jsvmapi.h
```diff
@@ -273,18 +252,46 @@ NODE_EXTERN napi_status napi_define_class(napi_env e,
...
-                                                  napi_value* result);
-
-NODE_EXTERN napi_status napi_create_weakref(napi_env e, napi_value v,
-                                            napi_weakref* result);
-NODE_EXTERN napi_status napi_get_weakref_value(napi_env e, napi_weakref w,
-                                               napi_value* result);
-NODE_EXTERN napi_status napi_release_weakref(napi_env e, napi_weakref w);
+
+// Set initial_refcount to 0 for a weak reference, >0 for a strong reference.
+NODE_EXTERN napi_status napi_create_reference(napi_env e,
```
digitalinfinity, line 272: - Are 0 and 1 the only expected initial_refcounts? If so, it might be nicer to just change initial refcount to an enum?
    - So can a reference that is created as weak be promoted to strong? How does this work with Chakra? (presumably Chakra's weak reference handles aren't used right, we just use AddRef/Release semantics).
      My vote is to prevent this from happening in the sense that if we increase the reference from 0 to 1, it should be an error but I could go either way

digitalinfinity:
Overall, looks in decent shape 👍 - couple comments left, I don't think they all need to be addressed before this PR is merged, but I would like some of the questions answered so that the model is clear in my head 😄

test/addons-abi/3_callbacks/binding.cc
jasongin: If you want convenience, use the C++ wrapper. That is what we should recommend for almost anyone using these APIs, as it makes everything much easier.

We have several APIs that take array and count parameters, and in all of them the count parameter is first. Elsewhere I've seen it both ways.

src/node_jsrtapi.cc
jasongin: This doesn't create an object with external data, it creates a value of type external. (In JSRT that happens to be stored internally as an Object, but in V8 it is not.)

test/addons-abi/7_factory_wrap/myobject.h
jasongin: The C++ wrapper does that.

src/node_jsrtapi.cc
jasongin: This `externalData` pointer is only ever allocated internally. The `_data` member holds the user's data, which may be freed by the provided callback field, `_cb`.

(The `ExternalData` class here is an internal implementation detail.)

src/node_jsrtapi.cc
jasongin: We could throw a JavaScript exception if it failed here, but it didn't seem necessary.

The constructed object will just have the default Object prototype. In most cases it won't matter unless someone does an `instanceof` check. (Wrapped classes cannot extend JS classes.)

src/node_jsrtapi.cc
jasongin: Yes. Normally it would be constructed via a NAPI function callback. If not, the call to `JsSetExternalData`, and therefore this function, will return an invalid-arg error.

I don't think it's worth representing it in the type system. It is sufficient to document that `napi_wrap` may only be used on the `this` parameter of a constructor callback function.

src/node_jsvmapi.h
jasongin: 0 and 1 would be most common, but >1 is possible and may avoid extra API calls if the caller knows they are adding  more references. Yes, references can be promoted from weak to strong, and demoted from strong to weak, by adjusting the refcount with `napi_reference_addref` and `napi_reference_release`. It works with Chakra using a combination of weak references and `AddRef`/`Release` on the target value.

src/node_jsvmapi_types.h
jasongin: An external is not an object. Think of it as just a `void*` that can be passed around as a JavaScript value. There is generally no need to attach external data to arbitrary JS Objects, and neither V8 nor Chakra can support that.

src/node_jsrtapi.cc
jasongin: If we don't call `JsAddRef` on the weak reference, it won't remain valid.

JSRT doesn't have a concept of "strong references", so they are emulated by calling `JsAddRef `on the target value.

Keep in mind NAPI references need to be able to promoted/demoted between weak and strong.

test/addons-abi/3_callbacks/binding.cc
digitalinfinity: Ok, I buy that. Still would be nice for a convenience macro to make reading some of this code easier but I don't care too strongly

src/node_jsrtapi.cc
digitalinfinity: Ah doh, yes- makes sense

src/node_jsvmapi.h
jasongin: Note async object-wrapping scenarios require promoting/demoting references. That is what `node::ObjectWrap`'s `Ref()` and `Unref()` methods are for.

src/node_jsrtapi.cc
```diff
@@ -1004,67 +1047,157 @@ napi_status napi_coerce_to_string(napi_env e, napi_value v, napi_value* result)
...
-  }
-  *result = reinterpret_cast<napi_persistent>(value);
-  return napi_ok;
-}
 
-napi_status napi_release_persistent(napi_env e, napi_persistent p) {
-  JsValueRef thePersistent = reinterpret_cast<JsValueRef>(p);
-  CHECK_JSRT(JsRelease(static_cast<JsRef>(thePersistent), nullptr));
+  CHECK_JSRT(JsCreateExternalObject(
+    new jsrtimpl::ExternalData(data, finalize_cb),
```
digitalinfinity: Is this a throwing new? Otherwise we should probably do the allocation separately and check for null?

src/node_jsrtapi.cc
digitalinfinity: Makes sense- what does typeof(obj) in JavaScript give for these objects? Is it different between v8 and Chakra?

src/node_jsrtapi.cc
jasongin: I think it is not. (I'll verify.) But yes I'll add a check for null to all the `new` calls.

src/node_jsvmapi.h
digitalinfinity: I'm still unclear how going from a weak reference to a strong reference works. So say you call release, and it's now at 0. At a later point, if you call AddRef to make the reference strong, what's stopping the GC from collecting the weak reference object itself in the interim?

src/node_jsrtapi.cc
jasongin: For Chakra it would be 'object'. For V8... I actually don't know.

They might be different, but it shouldn't matter because JavaScript code cannot interact with these kinds of values in any meaningful way.

src/node_jsrtapi.cc
digitalinfinity: Right- my question was, whether you needed to create the weak reference there in the first place, or whether you could use a different scheme (for instance- when you create a reference, you allocate an internal "handle" object, and maintain a map of strong references to handles- in addition, you could install a before collect callback that would null out the strong reference in the handle object- that way, you only have to maintain ref count on the strong reference itself, and the handle object would get destroyed when you call napi_delete_reference)- does that make sense?

src/node_jsvmapi.h
jasongin: Obviously nothing stops a weak-referenced value from being GC'd. If you try to increment the refcount after that, then the call to `napi_reference_addref` will return an error. But typically if you're using refcounting, you will start with a refcount of one, then increment/decrement until you get to 0.

digitalinfinity (approved):

src/node_jsvmapi.h
Fishrock123: Just a clarification: `ref()` and `unref()` is in relation to how libuv treats handles, not how GC works.

I'm not 100% sure about `node::ObjectWrap`, but all regular C++ handles are like that.

jasongin: @Fishrock123 a `napi_ref` is equivalent to a `v8::Persistent`, which can be a weak or strong reference to a JS value. [`node::ObjectWrap`'s `Ref()` and `Unref()` methods work by making the underlying `v8::Persistent` weak or strong](https://github.com/nodejs/node/blob/master/src/node_object_wrap.h#L72-L93).

jasongin: This is not related to libuv's `uv_ref()` and `uv_unref()` handle APIs.

src/node_jsrtapi.cc
jasongin: Discussed with @digitalinfinity offline. We agreed that this could be optimized as he suggested, and also that would eliminate the justification for the `JsCreateWeakReference()` API that was added to JSRT (in the NAPI branch) specifically for NAPI. This is an implementation detail that won't affect the public API surface, so we'll keep this change as it is now and add the optimization later.

PR #116, opened 2017-02-24
Enable custom tagged binary build

jasongin:
We need this change in the `Makefile` to enable custom-tagged builds for NAPI releases. Then when you run `node --version` with a NAPI build, it can output something like `v8.0.0-v8-napi-v0.1.0` or `v8.0.0-chakracore-napi-v0.1.0` rather than a nightly build tag.

From discussion with @aruneshchandra, as long as NAPI is a separate project from node we will version NAPI releases independently from the node version, because of how a single NAPI release includes builds of multiple node versions.

No equivalent change is needed for `vcbuild.bat` on Windows because it doesn't have this restriction for custom tagged builds.

I'm not sure whether this change would be included in an eventual PR to node. It seems generally correct to me to allow custom-tagged builds regardless of documentation status.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -549,7 +549,7 @@ PACKAGEMAKER ?= /Developer/Applications/Utilities/PackageMaker.app/Contents/MacO
 PKGDIR=out/dist-osx
 
 release-only:
-	@if [ "$(DISTTYPE)" != "nightly" ] && [ "$(DISTTYPE)" != "next-nightly" ] && \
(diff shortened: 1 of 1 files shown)


digitalinfinity:
LGTM. @mhdawson @nodejs/build is there any reason why custom tagged builds were not excluded in this condition in the first place? Would this change be acceptable to the mainline node master branch?

mhdawson:
Discussed in latest N-API meeting. Don't need this anymore so will close for now.

PR #117, opened 2017-02-26
Revamp buffer API (ChakraCore)

jasongin:
This is the ChakraCore equivalent of the buffer API changes implemented for V8 as #109 and discussed as #108.

Changes to test files are copied directly from the V8 branch, so those don't need to be re-reviewed.


boingoing:
Looks good to me.

gabrielschulhof (approved):
LGTM

PR #120, opened 2017-03-01
perf: Fix to nanomsg performance

sampsongao:


src/node_api_helpers.h
```diff
@@ -132,23 +132,13 @@ namespace Napi {
...
    public:
-    Callback() {
-      napi_env env;
-      napi_get_current_env(&env);
-      HandleScope scope(env);
-      napi_value obj;
-      napi_create_object(env, &obj);
-      napi_create_reference(env, obj, 1, &handle);
-    }
+    Callback() {}
```
mhdawson, line 135: I think you need to at least set handle to null here

src/node_api_helpers.h
```diff
@@ -205,30 +190,25 @@ namespace Napi {
       HandleScope scope;
       napi_env env;
       napi_get_current_env(&env);
-      napi_value h;
-      napi_get_reference_value(env, handle, &h);
-      napi_set_element(env, h, kCallbackIndex, fn);
+      napi_delete_reference(env, handle);
```
mhdawson, line 193: This probably needs a check that handle is not null

kkoopa: Only because of the `CHECK_ARG` in `napi_delete_reference` (is that actually necessary?). Calling `delete` on `nullptr` is perfectly fine.

src/node_api_helpers.h
kkoopa: Yes, `Callback() : handle{nullptr} {}`. Otherwise the pointer is uninitialized and although it might not be a problem right now, any small change which relies on the pointer being initialized would implode.

jasongin:
Note the entire `node-api-helpers.h` header file should be obsoleted by [abi-stable-node-wrapper](https://github.com/nodejs/abi-stable-node-wrapper). We should update our ported code to use that eventually; if there's any issue with that then it should be improved accordingly.

aruneshchandra:
Not needed anymore.

PR #125, opened 2017-03-02
napi: fix external constructor call in jsrt impl

digitalinfinity:
In the JSRT implementation, External Constructor calls were using undefined as the default return value if the external call didn't return anything.

Changed to set the this pointer as the default return value if the call was being invoked as a constructor


jasongin (approved):

boingoing (approved):
LGTM

jasongin:
Does V8 do this automatically? (We don't have code in `jsvmimpl.cc` that does it.)

PR #128, opened 2017-03-06
docs: vm summit 2017-03-03 minutes

joshgav:
cc @aruneshchandra @mhdawson @nodejs/api 

@jasnell should we collect these notes in https://github.com/nodejs/vm? This repo seems a subset of that effort.

### DIFF
--- a/meeting notes/2017-03-03.md
+++ b/meeting notes/2017-03-03.md
@@ -1,0 +1,353 @@
+# VM Summit #3 - 2017-03-03
+
+* Recording: https://www.youtube.com/watch?v=ZRVjWhBEwB0
+* Notes:
+  https://docs.google.com/document/d/1mIYQ6YjI2yUS07h5dtkXW-EKgyY2_XYjgVFaZt0Yagk
+
+# Attendees (in person):
+
+* James Snell @jasnell
+* Dan Shaw @dshaw
+* Josh Gavant @joshgav
+* Hitesh Kanwathirtha @digitalinfinity
+* Mark Marron @mrkmarron
+* Arunesh Chandra @aruneshchandra
+* Taylor Woll @boingoing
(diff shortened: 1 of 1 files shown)


jasnell:
hmm.. not sure really.

aruneshchandra (approved):

meeting notes/2017-03-03.md
```diff
@@ -0,0 +1,353 @@
...
+
+C++ API wrapper over C API for convenience, but not planning to include in core.
+
+V8 Isolate APIs little used in modules and don’t plan to include in N-API.
+[deasync](https://www.npmjs.com/package/deasync) uses many of these.
+
+Demo from Arunesh: https://github.com/boingoing/napi_demo. Same native modules -
+canvas, leveldown - used with 6, 8, and Node-ChakraCore.
+
+Canvas perf regression is likely because there’s heavy usage of V8 APIs so much
```
digitalinfinity, line 118: This is not quite accurate. Heavy usage of v8 APIs is not really a problem for NAPI- rather, the canvas perf regressions are likely because the canvas benchmarks that regress contain a trivial amount of real work compared to the v8 API calls they make, leading to the API overhead dominating the benchmark

digitalinfinity (approved):

meeting notes/2017-03-03.md
mhdawson: I'd write it as:
Canvas perf regression is likely because the performance benchmark makes calls to canvas where the work done in the native method is very small.  This means that any overhead, even the limited overhead added by N-API has a significant impact.

mhdawson (approved):
Other than my one suggestion for a wording change, LTGM.

mhdawson:
Replaced by https://github.com/nodejs/vm/pull/5#pullrequestreview-27408121

PR #129, opened 2017-03-07
Remove napi registration code from node.cc (V8)

jasongin:
I realized that it is possible to register NAPI modules without any NAPI-specific registration code in `node.cc`. This will make our diff much smaller in the eventual PR, and eliminate any possible concerns about a behavior change for "corrupt" modules having a version of -1.

Previously, `napi_module_register()` would simply forward the module struct unmodified to `node_module_register`. With this change it sets the module version to the current `NODE_MODULE_VERSION` (instead of -1). And it intercepts the registration callback in order to convert the `env`, `module` and `exports` objects to VM-agnostic types.

Also note NAPI is now using the "context-aware" module registration function (available in Node >= 4.0) to get the isolate from the context rather than calling `Isolate::GetCurrent()`.

Now the only NAPI-specific code in `node.cc` is for the command-line flag. (I think we plan to invert that, but I'll leave it for a later change.)

### DIFF
--- a/src/node.cc
+++ b/src/node.cc
@@ -2430,15 +2430,24 @@ void DLOpen(const FunctionCallbackInfo<Value>& args) {
 
   if (mp->nm_version != NODE_MODULE_VERSION) {
     char errmsg[1024];
-    snprintf(errmsg,
-             sizeof(errmsg),
-             "The module '%s'"
-             "\nwas compiled against a different Node.js version using"
-             "\nNODE_MODULE_VERSION %d. This version of Node.js requires"
-             "\nNODE_MODULE_VERSION %d. Please try re-compiling or "
-             "re-installing\nthe module (for instance, using `npm rebuild` or "
-             "`npm install`).",
-             *filename, mp->nm_version, NODE_MODULE_VERSION);
+    if (mp->nm_version == -1) {
+      snprintf(errmsg,
+               sizeof(errmsg),
+               "The module '%s'"
+               "\nwas compiled against the Node.js API. This feature is "
+               "\nexperimental and must be enabled on the command-line.",
+               *filename);
+    } else {
+      snprintf(errmsg,
+               sizeof(errmsg),
+               "The module '%s'"
+               "\nwas compiled against a different Node.js version using"
+               "\nNODE_MODULE_VERSION %d. This version of Node.js requires"
+               "\nNODE_MODULE_VERSION %d. Please try re-compiling or "
+               "re-installing\nthe module (for instance, using `npm rebuild` or "
+               "`npm install`).",
+               *filename, mp->nm_version, NODE_MODULE_VERSION);
+    }
 
     // NOTE: `mp` is allocated inside of the shared library's memory, calling
     // `uv_dlclose` will deallocate it
--- a/src/node.cc
+++ b/src/node.cc
@@ -1,4 +1,4 @@
-﻿#include "node.h"
+#include "node.h"
 #include "node_buffer.h"
 #include "node_constants.h"
 #include "node_file.h"
--- a/src/node.cc
+++ b/src/node.cc
@@ -1,4 +1,4 @@
-#include "node.h"
+﻿#include "node.h"
 #include "node_buffer.h"
 #include "node_constants.h"
 #include "node_file.h"
@@ -2428,13 +2428,7 @@ void DLOpen(const FunctionCallbackInfo<Value>& args) {
     return;
   }
 
-#ifdef ENABLE_NAPI
-  bool isNapiModule = (!no_napi_modules && mp->nm_version == -1);
-
-  if (mp->nm_version != NODE_MODULE_VERSION && !isNapiModule) {
-#else /* !defined ENABLE_NAPI */
   if (mp->nm_version != NODE_MODULE_VERSION) {
-#endif /* def ENABLE_NAPI */
     char errmsg[1024];
     snprintf(errmsg,
              sizeof(errmsg),
@@ -2465,21 +2459,6 @@ void DLOpen(const FunctionCallbackInfo<Value>& args) {
   Local<String> exports_string = env->exports_string();
   Local<Object> exports = module->Get(exports_string)->ToObject(env->isolate());
 
-#ifdef ENABLE_NAPI
-  if (isNapiModule) {
-    if (mp->nm_register_func != nullptr) {
-      reinterpret_cast<node::addon_abi_register_func>(mp->nm_register_func)(
-          v8impl::JsEnvFromV8Isolate(v8::Isolate::GetCurrent()),
-          v8impl::JsValueFromV8LocalValue(exports),
-          v8impl::JsValueFromV8LocalValue(module),
-          mp->nm_priv);
-    } else {
-      uv_dlclose(&lib);
-      env->ThrowError("Module has no declared entry point.");
-    }
-    return;
-  }
-#endif /* def ENABLE_NAPI */
   if (mp->nm_context_register_func != nullptr) {
     mp->nm_context_register_func(exports, module, env->context(), mp->nm_priv);
   } else if (mp->nm_register_func != nullptr) {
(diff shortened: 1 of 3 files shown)


src/node_jsvmapi.cc
```diff
@@ -496,6 +492,48 @@ namespace v8impl {
...
+  // Indicates whether NAPI was enabled/disabled via the node command-line.
+  extern bool no_napi_modules;
+}
+
+// Registers a NAPI module.
+void napi_module_register(napi_module* mod) {
+  // NAPI modules always work with the current node version.
+  int moduleVersion = NODE_MODULE_VERSION;
+  if (node::no_napi_modules) {
+    // NAPI is dsabled, so set the module version to -1 to cause the module to be unloaded.
```
boingoing: nitpick: typo in disabled

src/node_jsvmapi.cc
```diff
@@ -496,6 +492,48 @@ namespace v8impl {
...
+    moduleVersion = -1;
+  }
+
+  static node::node_module nm = {
+    moduleVersion,
+    mod->nm_flags,
+    nullptr,
```
boingoing: Any chance a module would want to use these nullptr fields?

jasongin:
CI: https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-commit-linux-abi/9/

jasongin:
Also fixing https://github.com/nodejs/abi-stable-node/issues/133

src/node.cc
```diff
@@ -1,4 +1,4 @@
-#include "node.h"
+﻿#include "node.h"
```
gabrielschulhof: Does this change add or remove that non-printing character? If it adds it, please drop this.

src/node_jsvmapi.cc
gabrielschulhof: disabled

src/node_jsvmapi.h
```diff
@@ -1,4 +1,4 @@
-/*******************************************************************************
+﻿/*******************************************************************************
```
gabrielschulhof: Again, if this change introduces a non-printing character, please drop it.

gabrielschulhof (commented):
Looking good, aside from a few whitespace nits.

src/node_jsvmapi.cc
```diff
@@ -1,4 +1,4 @@
-/*******************************************************************************
+﻿/*******************************************************************************
```
gabrielschulhof: Please drop this change as well!

jasongin:
@gabrielschulhof Those are just standard UTF-8 byte-order-marks. They're technically required if the file has any non-ASCII characters, but since these files don't then I don't know why my editor added them. They shouldn't cause any problems in any tools or editors on any platform, but I'll remove them anyway.

src/node_jsvmapi.cc
```diff
@@ -496,6 +492,48 @@ namespace v8impl {
 
 }  // end of namespace v8impl
 
+// Intercepts the Node-V8 module registration callback. Converts parameters to NAPI equivalents
+// and then calls the registration callback specified by the NAPI module.
+void napi_module_register_cb(v8::Local<v8::Object> exports,
+                             v8::Local<v8::Value> module,
+                             v8::Local<v8::Context> context,
+                             void* priv) {
+  napi_module* mod = static_cast<napi_module*>(priv);
```
digitalinfinity, line 501: is there any error checking needed here? Can any of the calls below fail? if so, what happens?

src/node_jsvmapi.cc
```diff
@@ -496,6 +492,48 @@ namespace v8impl {
...
+  extern bool no_napi_modules;
+}
+
+// Registers a NAPI module.
+void napi_module_register(napi_module* mod) {
+  // NAPI modules always work with the current node version.
+  int moduleVersion = NODE_MODULE_VERSION;
+  if (node::no_napi_modules) {
+    // NAPI is dsabled, so set the module version to -1 to cause the module to be unloaded.
+    moduleVersion = -1;
```
digitalinfinity: Is this explicitly needed or is it sufficient to just return early here?

digitalinfinity (approved):
LGTM, just a couple questions

src/node_jsvmapi.cc
jasongin: The `v8impl::` calls below are merely casts that evaporate in compilation.
The `nm_register_func()` returns `void`, so there's nothing to check.

Obviously if `mod` or `context` are null then this could crash, but that would never happen unless something is very broken in the node engine. I don't see anything worth checking.

src/node_jsvmapi.cc
jasongin: If this code returned early then node would complain with the error message: `Module did not self-register.`
By attempting to register with an invalid version (-1), the error message is:
```
The module '%s'
was compiled against a different Node.js version using
NODE_MODULE_VERSION -1. This version of Node.js requires
NODE_MODULE_VERSION 52. Please try re-compiling or re-installing
the module (for instance, using `npm rebuild` or `npm install`).
```
That's not quite ideal, but it's better than the other message.

src/node_jsvmapi.cc
digitalinfinity: Ok- casts should be fine- I wasn't sure if there was any memory allocation happening here, hence the comment

src/node_jsvmapi.cc
digitalinfinity: Agreed that the error message is better than the self-registration failure message- curious though, should we improve on this error message?

jasongin: Yes, we could easily show a different error message when the module version is -1. Nobody should be concerned about seeing that in our PR diff. I'll add that.

src/node_jsvmapi.h
```diff
@@ -31,20 +31,14 @@ NODE_EXTERN typedef void (*addon_abi_register_func)(
   void* priv);
 }  // namespace node
 
-struct napi_module_struct {
-  int nm_version;
```
mhdawson: It may be prudent to add a napi version here, its easier to simply leave it at 1.0 than have to add an entry to the struct later on.

Thinking about it we may want to block off from private values at the end so that if we do need to expand later on we can do that as otherwise any change to this structure would for us into what we needed to do for 0.10.X with additional lookups.

I guess I'm suggested we add a version, and then say 4 void* fields at the end that we might use later on.

mhdawson (approved):
Other than my one suggestion for possible additions to the struct LGTM.

jasongin:
@mhdawson @digitalinfinity I added a commit with napi-specific error message and version field in the module struct. Do you want to review the update?

src/node_jsvmapi.cc
```diff
@@ -496,6 +492,48 @@ namespace v8impl {
...
+  static node::node_module nm = {
+    moduleVersion,
+    mod->nm_flags,
+    nullptr,
+    mod->nm_filename,
+    nullptr,
+    napi_module_register_cb,
+    mod->nm_modname,
+    mod, // priv
+    nullptr,
```
digitalinfinity, line 532: noob c question - does this set all 4 void*'s to null or just the first, and leaves the rest uninitialized? Is this compiler specific behavior or per the standard?

digitalinfinity (approved):
LGTM with one question

src/node_jsvmapi.cc
jasongin: `{0}` initializes the array to all zeros. See **Notes** at http://en.cppreference.com/w/c/language/array_initialization

boingoing (approved):
This looks great. NAPI_MODULE_VERSION should be useful for the NAPI versioning story.

src/node_jsvmapi.cc
jasongin: Already fixed. Please look at the latest commit.

boingoing: I added this comment against the earlier commit but I guess it didn't post because it was waiting for me to complete the review. Didn't realize it did that. Disregard, then.

jasongin: My mistake. I did see it before (which is why I fixed it). But then I was confused by how GitHub posted the comment after your signoff.

src/node_jsvmapi.cc
jasongin: No, I don't think so. The 3 `nullptr` fields are:
 * `nm_dso_handle` - used internally by node to store the uv module handle; it is never provided via the module registration func.
 * `nm_register_func` - older signature of the register func (`addon_register_func`). We don't need to use it because we use the newer signature (`addon_context_register_func`) that takes the V8 context (and isolate) as a parameter.
 * `nm_link` - used internally by node to create a linked-list of modules; it is never provided by the module registration func.

jasongin: Hmm, I just realized the static-ness is going to be a problem because registering a second module will overwrite the first: it looks like node doesn't make a copy of the registration struct. We'll need to allocate this on the heap instead.

mhdawson (approved):
ok LGTM with all the changes.

PR #131, opened 2017-03-07
Distinguish external value type from wrapped (ChakraCore)

jasongin:
This fixes a bug in the JSRT implementation of the NAPI "external" value type. Because JSRT doesn't actually have an external type (like V8 does), an external value is represented as an object with external data. But there needs to be some way to distinguish external values from wrapped objects, which also have external data. The solution is to set the external value object as not-extensible. That more closely matches the V8 implementation anyway because V8 external values cannot have properties.

This change includes a test case that failed before the fix and now passes.

### DIFF
--- a/src/node_jsrtapi.cc
+++ b/src/node_jsrtapi.cc
@@ -726,9 +726,19 @@ napi_status napi_get_type_of_value(napi_env e, napi_value vv, napi_valuetype* re
     case JsSymbol: *result = napi_symbol; break;
 
     default:
+      // An "external" value is represented in JSRT as an Object that has external data and
+      // DOES NOT allow extensions. (A wrapped object has external data and DOES allow extensions.)
       bool hasExternalData;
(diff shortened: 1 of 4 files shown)


jasongin:
The CI doesn't currently work for chakracore builds: it omits the engine=chakracore parameter to ./configure.

boingoing (approved):
Looks good to me. 

Do you think we will ever want to expose JsPreventExtension to module code via napi_freeze_object or something like that? In that case we wouldn't always be able to rely on this check, right?

jasongin:
It's possible (since V8 provides that functionality) but I think unlikely that we'd ever add an API for that.

It's always possible to call `Object.preventExtensions()` in JavaScript. And doing that on a wrapped object would make NAPI treat it as an external... but that probably will never be an issue in practice, and I can't think of a good alternative implementation.

PR #132, opened 2017-03-07
src: fix up cpplint errors in node_jsvmapi.h

mhdawson:
Starting to work down the list of errors we need correct
before submitting PR to core

### DIFF
--- a/src/node_jsvmapi.h
+++ b/src/node_jsvmapi.h
@@ -93,103 +93,170 @@ NODE_EXTERN napi_status napi_get_global(napi_env e, napi_value* result);
 NODE_EXTERN napi_status napi_create_object(napi_env e, napi_value* result);
 NODE_EXTERN napi_status napi_create_array(napi_env e, napi_value* result);
 NODE_EXTERN napi_status napi_create_array_with_length(napi_env e, int length,
-                                                        napi_value* result);
-NODE_EXTERN napi_status napi_create_number(napi_env e, double val, napi_value* result);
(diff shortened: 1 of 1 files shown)


mhdawson:
CI run: https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-commit-linux-abi/

src/node_jsvmapi.h
```diff
@@ -203,42 +270,54 @@ NODE_EXTERN napi_status napi_make_callback(napi_env e,
...
+                                                napi_callback_info cbinfo,
+                                                int* result);
+NODE_EXTERN napi_status napi_get_cb_args(napi_env e,
+                                         napi_callback_info cbinfo,
+                                         napi_value* buffer,
+                                         int bufferlength);
+NODE_EXTERN napi_status napi_get_cb_this(
+                                napi_env e,
+                                napi_callback_info cbinfo,
+                                napi_value* result);
```
jasongin: This one is incorrectly aligned.

jasongin (approved):

src/node_jsvmapi.h
mhdawson: I'll double check but I had to do that in several cases to avoid hitting the 80 char limit

mhdawson: Looks like that one can be aligned, fixing.

mhdawson:
Pushed commit to address comments so far.  Will squash when landing.

mhdawson:
Landed as 47b3ca0223c691273bd4a3614a3c9a20de7b8d2d

Issue #137, opened 2017-03-09
Remove build flag


gabrielschulhof:
... but keep the runtime flag, and maybe rename it to something like `--napi-modules=yes|no` instead of `--with|without-napi-modules` so we don't need to rename the option later.

mhdawson:
agree we should rename to --napi-modules

aruneshchandra:
Done.

PR #143, opened 2017-03-14
napi: remove build flag but keep runtime flag

gabrielschulhof:
Fixes https://github.com/nodejs/abi-stable-node/issues/137

I haven't yet tested on Windows whether it refuses to load a N-API module with `--napi-modules=no`.

The patch is organized as an intentional bug. It's set up to not allow N-API modules by default, but the global`load_napi_modules` in node.cc is initially set to `true`. When we submit the PR, we need to set that `false` first. Then, it will be necessary to specify either `--napi-modules` or `--napi-modules=yes` on the node command line.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,6 @@ FLAKY_TESTS ?= run
 TEST_CI_ARGS ?=
 STAGINGSERVER ?= node-www
 LOGLEVEL ?= silent
-NAPI ?= True
 OSTYPE := $(shell uname -s | tr '[A-Z]' '[a-z]')
 
 ifdef JOBS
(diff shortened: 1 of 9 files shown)


boingoing:
LGTM. 

After we switch the global ```load_napi_modules``` to false, will we need to add the flag to the command line for tests in addons-abi?

gabrielschulhof:
That's a good point. I should change the appropriate files to correctly
invoke node for the addons-abi tests.

mhdawson:
ok will wait for additions to addons-abi tests before reviewing.   Is there any easy diff against master with this change as its the remaining delta thats interesting.

gabrielschulhof:
@mhdawson
https://github.com/nodejs/abi-stable-node/compare/1b30df10037d24224a555829754f4423d35bbf88...gabrielschulhof:137-remove-build-flag (click on "Files Changed")

1b30df10037d24224a555829754f4423d35bbf88 is the last commit before we diverged, so that should always be the one used on the left hand side of the `...`

gabrielschulhof:
https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-commit-linux-abi/29/

mhdawson (approved):
LGTM

mhdawson:
CI green as well :)

Issue #135, opened 2017-03-08
Buffer data argument should be void*, not char*


kkoopa:
As I mentioned in [another issue](https://github.com/nodejs/abi-stable-node-wrapper/pull/1#issuecomment-283436327), it is arguably a historical oversight that `node::Buffer` takes a `char*` as the `data` argument and not a `void*`. This leads to mandatory typecasting whenever something else than a `char*` is given to `node::Buffer`. Now is a good chance not to repeat the same mistake.

```c++
NODE_EXTERN napi_status napi_create_buffer(napi_env e,
                                           size_t size,
                                           void** data,
                                           napi_value* result);
NODE_EXTERN napi_status napi_create_external_buffer(napi_env e,
                                                    size_t size,
                                                    void* data,
                                                    napi_finalize finalize_cb,
                                                    napi_value* result);
NODE_EXTERN napi_status napi_create_buffer_copy(napi_env e,
                                                const void* data,
                                                size_t size,
                                                napi_value* result);
NODE_EXTERN napi_status napi_is_buffer(napi_env e, napi_value v, bool* result);
NODE_EXTERN napi_status napi_get_buffer_info(napi_env e,
                                             napi_value v,
                                             void** data,
                                             size_t* length);
```

Additionally, the C++ layer could use a class template parametrized on the type of `data` to ensure further type safety.

boingoing:
Resolved by #145

PR #145, opened 2017-03-15
napi: Buffer data argument should be void* instead of char*

boingoing:
Seems the fact that these API take ```char*``` for the data parameter was more of a historical oversight in the design of ```node::Buffer``` than anything else. We should take the opportunity to clean up this rough edge with the new API contract.

Fixes https://github.com/nodejs/abi-stable-node/issues/135

### DIFF
--- a/src/node_jsvmapi.cc
+++ b/src/node_jsvmapi.cc
@@ -1,4 +1,4 @@
-/*******************************************************************************
+﻿/*******************************************************************************
  * Experimental prototype for demonstrating VM agnostic and ABI stable API
  * for native modules to use instead of using Nan and V8 APIs directly.
  *
@@ -2020,7 +2020,7 @@ napi_status napi_get_and_clear_last_exception(napi_env e, napi_value* result) {
   return napi_ok;
 }
 
-napi_status napi_create_buffer(napi_env e, size_t size, char** data,
+napi_status napi_create_buffer(napi_env e, size_t size, void** data,
                                napi_value* result) {
   NAPI_PREAMBLE(e);
   CHECK_ARG(data);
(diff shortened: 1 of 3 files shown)


test/addons-abi/test_buffer/test_buffer.cc
```diff
@@ -1,4 +1,4 @@
-#include <string.h>
+﻿#include <string.h>
```
mhdawson, line 1: what is the change on this line ?

src/node_jsvmapi.h
```diff
@@ -1,4 +1,4 @@
-/*******************************************************************************
+﻿/*******************************************************************************
```
mhdawson, line 1: what is the change here ?

src/node_jsvmapi.cc
```diff
@@ -1,4 +1,4 @@
-/*******************************************************************************
+﻿/*******************************************************************************
```
mhdawson, line 1: what is the change here ?

mhdawson (approved):
LGTM provided we either remove or explain the white space change.

mhdawson:
CI run: https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-commit-linux-abi/25/

test/addons-abi/test_buffer/test_buffer.cc
jasongin: I think this is a bug in VS 2017, which Taylor and I have started using more lately. VS 2017 apparently always adds a UTF-8 byte order mark to a source file when editing it even if there are no non-ASCII characters.

I think technically the BOM is required if there are non-ASCII characters in the file, and optional if the file is all ASCII. But VS shouldn't add the BOM when it wasn't there before and isn't necessary.

boingoing: Yeah I noticed this, too. Is it worth removing all the BOM characters?

PR #146, opened 2017-03-15
Argv array pointers should be const

jasongin:
I noticed this bug when I was working on the C++ wrappers and had to use `const_cast` in a few places where it shouldn't have been necessary.


boingoing:
LGTM 👍

PR #150, opened 2017-03-16
napi: Remove napi_propertyname

boingoing:
This change removes ```napi_propertyname``` and replaces their use with ```napi_value```. On v8 ```napi_propertyname``` is just a ```v8::Local<v8::String>``` anyway. It doesn't seem to be worthwhile from the types of code patterns we've seen in modules to keep the separate type for property keys.

Also introduces property get/set/has for properties named directly by a ```const char*``` which seems to be the largest use of property access on objects. Naming the property key via a utf8 string is keeping more in-line with the other napi method calls for dealing with names of things like ```napi_property_descriptor``` or the ```const char*``` we use to name classes (via ```napi_define_class```) or functions (via ```napi_create_function```).

Adds a few simple tests of these APIs.

### DIFF
--- a/src/node_jsvmapi.cc
+++ b/src/node_jsvmapi.cc
@@ -1,4 +1,4 @@
-/******************************************************************************
+﻿/******************************************************************************
  * Experimental prototype for demonstrating VM agnostic and ABI stable API
  * for native modules to use instead of using Nan and V8 APIs directly.
  *
@@ -72,17 +72,6 @@ V8EscapableHandleScopeFromJsEscapableHandleScope(
   return reinterpret_cast<EscapableHandleScopeWrapper*>(s);
 }
 
-v8::Local<v8::String> V8LocalStringFromJsPropertyName(napi_propertyname pn) {
-  // Likewise awkward
-  union U {
-    napi_propertyname pn;
-    v8::Local<v8::String> l;
-    U(napi_propertyname _pn) : pn(_pn) { }
-  } u(pn);
-  assert(sizeof(u.pn) == sizeof(u.l));
-  return u.l;
-}
-
 //=== Conversion between V8 Handles and napi_value ========================
 
 // This is assuming v8::Local<> will always be implemented with a single
@@ -108,17 +97,6 @@ v8::Local<v8::Value> V8LocalValueFromJsValue(napi_value v) {
     U(napi_value _v) : v(_v) {}
   } u(v);
   assert(sizeof(u.v) == sizeof(u.l));
-  return u.l;
-}
-
-v8::Local<v8::Value> V8LocalValueFromJsPropertyName(napi_propertyname pn) {
-  // Likewise awkward
-  union U {
-    napi_propertyname pn;
-    v8::Local<v8::Value> l;
-    U(napi_propertyname _pn) : pn(_pn) {}
-  } u(pn);
-  assert(sizeof(u.pn) == sizeof(u.l));
   return u.l;
 }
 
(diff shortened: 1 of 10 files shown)


src/node_jsvmapi.h
```diff
@@ -188,34 +188,43 @@ NODE_EXTERN napi_status napi_coerce_to_string(napi_env e,
...
-                                           const char* utf8name,
-                                           napi_propertyname* result);
 NODE_EXTERN napi_status napi_get_propertynames(napi_env e,
                                                napi_value object,
                                                napi_value* result);
 NODE_EXTERN napi_status napi_set_property(napi_env e,
-                                          napi_value object,
-                                          napi_propertyname name,
+                                          napi_value o,
+                                          napi_value k,
```
jasongin: I'd prefer full words for the parameter names. It may not be entirely obvious that `o`=object, `k`=key, `v`=value. And it should be consistent with the rest of the declarations that mostly spell out parameter names, with the exception of the `e` parameter.

jasongin (approved):
LGTM aside from the parameter naming.

You should probably hold on to this until others have had a chance to review and discuss.

src/node_jsvmapi.cc
```diff
@@ -98,18 +98,7 @@ namespace v8impl {
...
-    union U {
-      napi_propertyname pn;
-      v8::Local<v8::String> l;
-      U(napi_propertyname _pn) : pn(_pn) { }
-    } u(pn);
-    assert(sizeof(u.pn) == sizeof(u.l));
-    return u.l;
-  }
-
+
```
gabrielschulhof: There are two unnecessary spaces in an otherwise blank line here.

src/node_jsvmapi.cc
```diff
@@ -785,52 +774,104 @@ napi_status napi_set_return_value(napi_env e,
...
+  v8::Maybe<bool> set_maybe = obj->Set(context, key, val);
+
+  RETURN_STATUS_IF_FALSE(set_maybe.FromMaybe(false), napi_generic_failure);
+  return GET_RETURN_STATUS();
+}
+
+napi_status napi_has_property(napi_env e, 
+                              napi_value o, 
+                              napi_value k,
```
gabrielschulhof: The above three lines also have trailing whitespace.

src/node_jsvmapi_types.h
```diff
@@ -1,4 +1,4 @@
-#ifndef SRC_NODE_JSVMAPI_TYPES_H_
+﻿#ifndef SRC_NODE_JSVMAPI_TYPES_H_
```
gabrielschulhof, line 1: There's a non-printing character being added here.

mhdawson (approved):
LGTM after fixing up the naming and the formatting comments. I was always on the fence with respect to having the separate type.

src/node_jsvmapi.h
boingoing: Agreed. I changed these names to match what's in the cc file - they were previously out-of-sync - but did consider changing them all to be full words. Didn't initially do this because most of the API in this file use letters for parameter names like this. 

I will make a pass and change these to use full words.

boingoing:
Went through and fixed up all the parameter names which seemed to be lacking description because they were single characters or shorthand. A couple still remain, but I think this is largely more readable now. Also tried to make sure the names of parameters are the same in both header and cc file.

jasongin (approved):
LGTM...
Please do a CI run before submitting.

boingoing:
Thanks, passing CI run is here:
https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-commit-linux-abi/42/

src/node_jsvmapi_types.h
boingoing: This is due to what looks like a bug in VS 2017 which always adds a BOM, even when it's not required. There are quite a few files with these marks so I propose we fix them all at once.

PR #161, opened 2017-03-17
Remove regfunc cast from NAPI_MODULE

jasongin:
NAPI has two variants of module registration macros: the one here for the C API (recently renamed to NAPI_MODULE) and [one for the C++ API](https://github.com/nodejs/abi-stable-node-wrapper/blob/master/napi-inl.h#L17) (to be renamed to... NODE_API_MODULE?). Because of this cast, if you use the wrong macro then the code can still compile, but at runtime you'll get a crash during registration, and the cause is not at all obvious even when debugging.

The cast originally was copied from the [similar macro in node.h](https://github.com/nodejs/abi-stable-node/blob/api-prototype-8.x/src/node.h#L430), but I'm pretty sure it's only there for historical reasons. (They added a parameter while remaining compatible with older code.) We don't need it for NAPI, and it is likely to cause confusion if we keep it.

Without the cast, you get a build error if you use the wrong macro, which is good.

### DIFF
--- a/src/node_api.h
+++ b/src/node_api.h
@@ -54,7 +54,7 @@ struct napi_module {
       NAPI_MODULE_VERSION,                                            \
       flags,                                                          \
       __FILE__,                                                       \
-      (node::napi_addon_register_func)regfunc,                        \
+      regfunc,                                                        \
       #modname,                                                       \
       priv,                                                           \
       {0},                                                            \
--- a/test/addons-napi/1_hello_world/binding.cc
+++ b/test/addons-napi/1_hello_world/binding.cc
@@ -12,7 +12,7 @@ void Method(napi_env env, napi_callback_info info) {
 #define DECLARE_NAPI_METHOD(name, func)                          \
   { name, func, 0, 0, 0, napi_default, 0 }
 
-void Init(napi_env env, napi_value exports, napi_value module) {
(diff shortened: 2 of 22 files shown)


jasongin:
Oh, hmm... It appears all the tests rely on that cast because they omit the `priv` parameter from their registration callback. Now I'm not sure what to do.

digitalinfinity:
Does it make sense to just update the tests to require their callback to conform to napi_addon_register_func's signature? Function pointer casting can otherwise lead to trouble down the road

jasongin:
Yes, that's probably best. I'll update the PR.

jasongin:
Pushed another commit to update all the test `Init()` functions to include the `void* priv` parameter that is required by `napi_addon_register_func`.

jasongin:
CI passed: https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-commit-linux-abi/46/

mhdawson (approved):
LGTM.  Agreed updated tests and removing is the best way to go.

mhdawson:
Landed the PR to fix existing lint errors and updated CI job to run lint step. This PR should therefore be rebased and the CI re-run to validate

jasongin:
CI passed after rebase: https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-commit-linux-abi/50/

Issue #162, opened 2017-03-17
Add warning message for --napi-modules 


aruneshchandra:

jasongin:
Done as https://github.com/nodejs/abi-stable-node/commit/7e2cbb944c07e25298a30a5372c3d01c79314a43

PR #163, opened 2017-03-17
Add warning about experimental status

jasongin:
Fixes https://github.com/nodejs/abi-stable-node/issues/162

Note the warning does not display when `load_napi_modules` *defaults* to on, which is currently the case. We planned to switch that default to off anyway when making the PR.

### DIFF
--- a/src/node.cc
+++ b/src/node.cc
@@ -3682,10 +3682,9 @@ static void ParseArgs(int* argc,
       force_repl = true;
     } else if (strcmp(arg, "--no-deprecation") == 0) {
       no_deprecation = true;
-    } else if (strcmp(arg, "--napi-modules") == 0 ||
-               strcmp(arg, "--napi-modules=yes") == 0) {
-      fprintf(stderr, "Warning: N-API is an experimental feature "
-        "and could change at any time.\n");
+    } else if (strcmp(arg, "--napi-modules") == 0) {
+      load_napi_modules = true;
+    } else if (strcmp(arg, "--napi-modules=yes") == 0) {
       load_napi_modules = true;
     } else if (strcmp(arg, "--napi-modules=no") == 0) {
       load_napi_modules = false;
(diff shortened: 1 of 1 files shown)


jasongin:
CI: https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-commit-linux-abi/51/

src/node.cc
```diff
@@ -3682,9 +3682,10 @@ static void ParseArgs(int* argc,
       force_repl = true;
     } else if (strcmp(arg, "--no-deprecation") == 0) {
       no_deprecation = true;
-    } else if (strcmp(arg, "--napi-modules") == 0) {
-      load_napi_modules = true;
-    } else if (strcmp(arg, "--napi-modules=yes") == 0) {
+    } else if (strcmp(arg, "--napi-modules") == 0 ||
+               strcmp(arg, "--napi-modules=yes") == 0) {
+      fprintf(stderr, "Warning: N-API is an experimental feature "
+        "and could change at any time.\n");
```
jasnell: hmm... I'd personally prefer a `process.emitWarning()` emitted right after bootstrap

jasongin: I don't understand... `process.emitWarning()` is a JavaScript API. From where would it be called?

This was modeled after the [similar warning for the experimental tracing feature](https://github.com/nodejs/node/commit/ba4847e879424ad173289e8fb96cc86a09ee899b#diff-cd53544f44aab2c697bcd7b6a57f23ccR4483).

jasongin: Oh I see there is a `ProcessEmitWarning()` function in `node.cc`. I'd just need to find a good place to call it... command-line parsing is too early.

jasongin: Pushed a commit that uses `ProcessEmitWarning()` instead.

boingoing (approved):
LGTM

digitalinfinity (approved):

PR #167, opened 2017-03-18
Enable back-compat with pre-N-API node

jasongin:
With a few small changes, we can use these same `node_api_*` source files for back-compat all the way down to node v4.0.

 - Instead of `NODE_EXTERN`, define a new `NAPI_EXTERN` macro with different logic that also enables static-linking N-API when targeting older node versions,
 - Now that we're not using `NODE_EXTERN`, I reverted the refactoring of `node_macros.h`, so the PR diff will be smaller.
 - V8 in node v4 does not have a `Persistent::Get()` method. That is trivial to work around because there is a `Local<Value>` constructor that takes a persistent and yields exactly the same thing.
 - V8 in node v4 does not have a `Persistent::SetWeak()` method that doesn't take a callback. This is more difficult to work around. (Passing `nullptr` for the callback is not allowed.) It's a problem for the `napi_wrap()` implementation for a complicated reason explained in https://github.com/nodejs/abi-stable-node/pull/118. Due to this limitation, as far as I can see it is not possible to exactly implement the previous behavior of `napi_wrap()` for node v4. Since we don't want different behavior on different Node versions, I think it's better to make `napi_wrap()` a little more restrictive - see the comments there. It's still good though because it works with the two main wrapping scenarios I know of: wrapping a singleton instance that lives forever (as leveldown does) where the reference is not tracked, and wrapping a refcounted instance (as the C++ `ObjectWrap` helper does, which is used by node-canvas and many others).

A related PR for the `node-api` repo uses a copy of the `node_api*` source files to enable back-compat: https://github.com/nodejs/node-api/pull/7


jasongin:
CI passed: https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-commit-linux-abi/54/

gabrielschulhof:
We also need to modify addons' `binding.gyp` files to also include node_api.cc - conditionally, if we only want forward compatibility, and unconditionally if we want forward and backwards compatibilty. In the latter case, the API will always be compiled into the addon.

Also, if we want to build against an older version of node, we need to ship node_api.cc and node_api_async.cc in a separate npm package, right?

So, can we say that a portion of the V8 API sufficient for implementing N-API *has* been ABI-stable since Node.js version 4.0?

LGTM, but please test this with a module such that you build it once and then run it unchanged with a version 4.x, 6.x, upstream master, and our api-prototype-8.x on at least Windows and Linux!

mhdawson:
How does this relate to to the Node object wrap functionality which we use previously and is used today in modules? 

The key thing I'd like you to confirm (my assumption is yes) is that even with the new restrictions, the N-API object wrap provides the same or more function that the existing node implementation.

mhdawson:
@gabrielschulhof I don't follow your discussion on getting backward compatibility.  As soon as the node_api.cc is compiled into the module, the module is tied to that specific node version as it is using the v8 shape which is tied to that version of Node.

jasongin:
> We also need to modify addons' binding.gyp files to also include node_api.cc - conditionally, if we only want forward compatibility, and unconditionally if we want forward and backwards compatibilty. In the latter case, the API will always be compiled into the addon.
Also, if we want to build against an older version of node, we need to ship node_api.cc and node_api_async.cc in a separate npm package, right?

Yes, all that is handled in the `node-api` repo PR linked above.

> So, can we say that a portion of the V8 API sufficient for implementing N-API has been ABI-stable since Node.js version 4.0?

It has at least been source-stable. I'm not sure it has been ABI-stable.

jasongin:
> The key thing I'd like you to confirm (my assumption is yes) is that even with the new restrictions, the N-API object wrap provides the same or more function that the existing node implementation.

Yes. The `Napi::ObjectWrap` class in the C++ helpers still provides exactly the same functionality as the `node::ObjectWrap` class.

The only scenario that could have worked before but doesn't work now would be if you called `napi_wrap` to get a reference to the wrapper object, then deleted the reference *before* the finalizer callback was invoked: now there could be a memory leak because the finalizer will not run after the reference is destroyed. But that's not really a valid scenario: the Node and NAPI `ObjectWrap` implementation always delete the reference in response to the finalizer callback.

jasongin:
> As soon as the node_api.cc is compiled into the module, the module is tied to that specific node version as it is using the v8 shape which is tied to that version of Node.

Yes, just as it has always been when compiling native modules. The "backward compatibility" aspect is that you get to use the NAPI APIs on older versions of node, but *without* the benefit of not having to rebuild across versions.

gabrielschulhof:
@mhdawson Sorry! I keep forgetting the mangling that C++ does.

mhdawson (approved):
LGTM

boingoing (approved):
LGTM

digitalinfinity (approved):

Issue #168, opened 2017-03-20
CI run broken by N-API experimental warning


mhdawson:
Looks like the CI runs are broken: https://ci.nodejs.org/job/node-test-commit-linux-abi/nodes=ubuntu1404-64/55/console

An example failure:

```
not ok 57 parallel/test-child-process-custom-fds
  ---
  duration_ms: 0.138
  severity: fail
  stack: |-
    /home/iojs/build/workspace/node-test-commit-linux-abi/nodes/ubuntu1404-64
    (node:31351) [DEP0006] DeprecationWarning: child_process: options.customFds option is deprecated. Use options.stdio instead.
    (node:31351) Warning: N-API is an experimental feature and could change at any time.
    
    assert.js:81
      throw new assert.AssertionError({
      ^
    AssertionError: 'Warning' === 'DeprecationWarning'
        at process.on.exports.mustCall (/home/iojs/build/workspace/node-test-commit-linux-abi/nodes/ubuntu1404-64/test/common.js:607:12)
        at process.<anonymous> (/home/iojs/build/workspace/node-test-commit-linux-abi/nodes/ubuntu1404-64/test/common.js:473:15)
        at emitOne (events.js:122:20)
        at process.emit (events.js:212:7)
        at internal/process/warning.js:74:13
        at _combinedTickCallback (internal/process/next_tick.js:73:7)
        at process._tickCallback (internal/process/next_tick.js:104:9)
  ...
```

So I believe it would be related to: https://github.com/nodejs/abi-stable-node/commit/7e2cbb944c07e25298a30a5372c3d01c79314a43

mhdawson:
Possibly related to the change which was made after the CI run in the PR.

mhdawson:
@jasongin

jasongin:
Sorry about that, I guess I forgot to run CI again after updating the PR.

We actually do want this warning here. I think I'll have to update all the tests to expect that N-API warning.

jasongin:
The failures were coming from non-NAPI tests that were expecting specific warnings but failed due to this unexpected warning. The best way to fix this is to just switch N-API to be disabled by default (as planned anyway), so that those other tests don't enable N-API and don't see this warning.

Fixed by https://github.com/nodejs/abi-stable-node/pull/169

PR #169, opened 2017-03-20
Disable N-API by default

jasongin:
Set N-API to be off by default, unless enabled via the `--napi-modules` command-line option.

Fixes https://github.com/nodejs/abi-stable-node/issues/168

CI: https://ci.nodejs.org/job/node-test-commit-linux-abi/56/


gabrielschulhof:
LGTM 👍

PR #171, opened 2017-03-21
doc: initial version of memory management documentation

mhdawson:


napi_memory_management.md
```diff
@@ -0,0 +1,131 @@
+# Memory Management
+
+As N-API calls are made, references to objects in the heap for the underlying
+vm may be returned as ```napi_values```.  These references must hold the
```
gabrielschulhof: Perhaps we should say "may be returned as variables of type &#96;napi_value&#96;", so that the item in backquotes is exactly the name of the type.

napi_memory_management.md
```diff
@@ -0,0 +1,131 @@
...
+objects 'live' until they are no longer required by the native code,
+otherwise the objects could be collected before the native code was
+finished using them.
+
+As objects references are returned they are associated with a
```
gabrielschulhof: object~~s~~

napi_memory_management.md
```diff
@@ -0,0 +1,131 @@
...
+of the native method call. The result is that, by default, references
+remain valid and the objects associated with these references will be
+held live for the lifespan of the native method call.   
+
+In many cases, however, it is necessary that the references remain valid for
+either a shorter or longer lifespan than that of the native method.
+The sections which follow describe the N-API functions than can be used
+to change the reference lifespan from the default.
+
+## Making reference lifespan shorter than that of native method
```
gabrielschulhof: of *the* native method

napi_memory_management.md
```diff
@@ -0,0 +1,131 @@
...
+
+```
+for (int i = 0; i < 1000000; i++) {napi_
+  napi_handle_scope scope;
+  napi_open_handle_scope(env, &scope);
+  napi_value result;
+  napi_get_element(e object, i, &result);
+  // do something with element
+  napi_close_handle_scope(env, scope);
+}
```
gabrielschulhof: The examples should take into account error handling. What if a N-API call returns non-`napi_ok`?

napi_memory_management.md
```diff
@@ -0,0 +1,131 @@
...
+```
+
+## Making reference lifespan longer than that of native method
+
+In some cases an addon will need to be able to create and reference objects
+with a lifespan longer than that of a single native method invocation. For
+example, if you want to create a constructor and later use that constructor
+in a request to creates instances, you will need to be able to reference
+the constructor object across many different instance creation requests. This
+would not be possible with a normal references returned as a napi_value as
```
gabrielschulhof: reference~~s~~

gabrielschulhof: &#96;napi_value&#96;

napi_memory_management.md
```diff
@@ -0,0 +1,131 @@
...
+described in the earlier section.  The lifespan or normal references are
```
gabrielschulhof: o*f* normal references...

gabrielschulhof: *is* managed by scopes

napi_memory_management.md
```diff
@@ -0,0 +1,131 @@
...
+managed by scopes and all scopes must be closed before the end of a native
+method.
+
+N-API provides methods to create persistent references.  Each persistent
+reference has an associated count with a value of 0 or higher. The count
+determines if the reference will keep the corresponding object live.  
+references with a count of 0 do not prevent the object from being collected,
```
gabrielschulhof: "references" must be capitalized here.

napi_memory_management.md
```diff
@@ -0,0 +1,131 @@
...
+and are often called 'weak' references. Any count greater than 0 will prevent
+the object from being collected.
+
+references can be created with an initial reference count.  The count can
+then be modified through ```napi_reference_addref()``` and
+```napi_reference_release()```.  If an object is collected while the count
+for a reference is 0, all subsequent calls to
+get the object associated with the reference (```napi_get_reference_value()```)
+will return NULL for the returned napi_value.  An attempt to call
```
gabrielschulhof: &#96;napi_value&#96;

napi_memory_management.md
```diff
@@ -0,0 +1,131 @@
...
+```napi_reference_addref()``` for a reference whose object has been collected,
```
gabrielschulhof: I don't think we need a comma after "collected"

napi_memory_management.md
```diff
@@ -0,0 +1,131 @@
...
+will result in an error.
+
+references must be deleted once they are no longer required by the addon. When
```
gabrielschulhof: "references" must be capitalized here.

napi_memory_management.md
```diff
@@ -0,0 +1,131 @@
...
+a reference is deleted it will no longer prevent the corresponding object from
+being collected.  Failure to delete a persistent reference will result in
+a 'memory leak' with both the native memory for the persistent reference and
+the corresponding object in heap being retained forever.
```
gabrielschulhof: in *the* heap ... although I believe is it customarily *on* the heap ...

napi_memory_management.md
```diff
@@ -0,0 +1,131 @@
...
+                                               napi_handle_scope* result);
+NODE_EXTERN napi_status napi_close_handle_scope(napi_env e,
+                                                napi_handle_scope scope);
+```
+
+N-API only supports a single nested hiearchy of scopes.  There is only one
+active scope at any time, and all new references will be associated with that
+scope while it is active.  Scopes must be closed in the reverse order from
+which they are opened.  In addition, all scopes created within a native method
+must be closed before returning from that method.
```
jasongin: This is missing discussion of *escapable* handle scopes. (We actually might want to simplify our API there. I opened https://github.com/nodejs/abi-stable-node/issues/177 to track that issue.)

mhdawson:
right I had thought about escapable scopes but it was getting late on the plane and I forgot.  Will look at #177 add in coverage of those to the doc.

napi_memory_management.md
mhdawson: I had excluded error handling on purpose to keep the examples short.  I've added in some minimal error checking

mhdawson:
Pushed updates to address @gabrielschulhof comments.  Will add description covering escapable scopes while on the plane later today.

napi_memory_management.md
```diff
@@ -0,0 +1,143 @@
...
+    napi_status status = napi_get_element(e object, i, &result);
+    if (status != napi_ok) {
+      break;
+    }
+    // do something with element
+  }
+```
+
+This would result in a large number of references being created, consuming
+substantial resources. In addition, even through the native code could only
```
gabrielschulhof: th~~r~~ough

gabrielschulhof: Being able to access only the last reference is equivalent to not being able to access the rest of the references, so there's a redundancy in this sentence.

gabrielschulhof: Perhaps: "...consuming substantial resources, because all the references in the loop would be kept alive as they all share the same scope, even though only the reference from the most recent iteration of the loop would actually be assigned to a variable".

napi_memory_management.md
```diff
@@ -0,0 +1,143 @@
...
+use the most recent reference, all of the associated objects would also be
+kept alive, even though the native method was no longer able to access them.
+
+To handle this case, N-API provides the ability to establish a new 'scope' to
```
gabrielschulhof: I'm interrupting the review here, and leaving this comment as a bookmark for myself. I'll delete it when I continue the review. BRB.

napi_memory_management.md
```diff
@@ -0,0 +1,143 @@
...
+the object from being collected.
+
+References can be created with an initial reference count.  The count can
+then be modified through ```napi_reference_addref()``` and
+```napi_reference_release()```.  If an object is collected while the count
+for a reference is 0, all subsequent calls to
+get the object associated with the reference (```napi_get_reference_value()```)
+will return NULL for the returned ```napi_value```.  An attempt to call
+```napi_reference_addref()``` for a reference whose object has been collected,
+will result in an error.
```
gabrielschulhof: BRB.

napi_memory_management.md
```diff
@@ -0,0 +1,143 @@
+# Memory Management
+
+As N-API calls are made, references to objects in the heap for the underlying
+vm may be returned as ```napi_value```.  These references must hold the
```
jasongin: Is there a reason you're using triple backticks for all inline code? Single backticks should be sufficient for formatting inline code; triple backticks are only necessary for fencing code blocks, as far as I know.

napi_memory_management.md
```diff
@@ -0,0 +1,143 @@
+# Memory Management
```
jasongin: Should this be titled something like "Object lifetime management" instead?

The only APIs we have that are directly related to memory-management are the finalize callbacks for `napi_wrap()`, `napi_create_external()`, `napi_create_external_buffer()`, and `napi_create_external_arraybuffer()` that enable freeing an external pointer, but those are not even covered here. (They belong in other topics.)

napi_memory_management.md
```diff
@@ -0,0 +1,143 @@
...
+
+As N-API calls are made, references to objects in the heap for the underlying
+vm may be returned as ```napi_value```.  These references must hold the
+objects 'live' until they are no longer required by the native code,
+otherwise the objects could be collected before the native code was
+finished using them.
+
+As object references are returned they are associated with a
```
jasongin: Should this doc use the term "handles" instead of "references"? Unless specifically referring to persistent or weak (`napi_ref`) references.

napi_memory_management.md
```diff
@@ -0,0 +1,143 @@
...
+N-API provides methods to create persistent references.  Each persistent
+reference has an associated count with a value of 0 or higher. The count
+determines if the reference will keep the corresponding object live.  
+References with a count of 0 do not prevent the object from being collected
+and are often called 'weak' references. Any count greater than 0 will prevent
+the object from being collected.
+
+References can be created with an initial reference count.  The count can
+then be modified through ```napi_reference_addref()``` and
+```napi_reference_release()```.  If an object is collected while the count
```
jasongin: These APIs were renamed to `napi_reference_ref()` and `napi_reference_unref()`.

mhdawson:
added section for escapable scopes.

Will look a new comments early next week.

napi_memory_management.md
mhdawson: fixed

napi_memory_management.md
mhdawson: Makes sense to me, changing.

napi_memory_management.md
mhdawson: That does fit better with the naming of the functions, updated.

mhdawson:
@gabrielschulhof @jasongin pushed update to address all comments.

jasongin (approved):

napi_memory_management.md
```diff
@@ -0,0 +1,175 @@
...
+      break;
+    }
+    // do something with element
+  }
+```
+
+This would result in a large number of handles being created, consuming
+substantial resources. In addition, even though the native code could only
+use the most recent handle, all of the associated objects would also be
+kept alive since they are share the same scope.
```
gabrielschulhof: ... they *all* share ...

napi_memory_management.md
```diff
@@ -0,0 +1,175 @@
...
+function:
+
+```
+NAPI_EXTERN napi_status napi_escape_handle(napi_env env,
+                                           napi_escapable_handle_scope scope,
+                                           napi_value escapee,
+                                           napi_value* result);
+```
+
+## References to objects with a lifespan longer than that of native method
```
gabrielschulhof: ... *the* native method

napi_memory_management.md
```diff
@@ -0,0 +1,175 @@
...
+References can be created with an initial reference count.  The count can
+then be modified through `napi_reference_ref()` and
+`napi_reference_unref()`.  If an object is collected while the count
+for a reference is 0, all subsequent calls to
+get the object associated with the reference (`napi_get_reference_value()`)
+will return NULL for the returned `napi_value`.  An attempt to call
+`napi_reference_ref()` for a reference whose object has been collected,
+will result in an error.
+
+references must be deleted once they are no longer required by the addon. When
```
gabrielschulhof: "references" should be capitalized.

mhdawson:
@gabrielschulhof pushed commit to address next set of comments.

mhdawson:
@gabrielschulhof ping

napi_memory_management.md
```diff
@@ -0,0 +1,175 @@
...
+held live for the lifespan of the native method call.   
+
+In many cases, however, it is necessary that the handles remain valid for
+either a shorter or longer lifespan than that of the native method.
+The sections which follow describe the N-API functions than can be used
+to change the handle lifespan from the default.
+
+## Making handle lifespan shorter than that of the native method
+
+It is often necessary to make the lifespan for handles shorter than
```
gabrielschulhof: "... lifespan *of* handles ... " maybe - no biggie.

napi_memory_management.md
```diff
@@ -0,0 +1,175 @@
...
+  }
+```
+
+This would result in a large number of handles being created, consuming
+substantial resources. In addition, even though the native code could only
+use the most recent handle, all of the associated objects would also be
+kept alive since they all share the same scope.
+
+To handle this case, N-API provides the ability to establish a new 'scope' to
+which newly created handle will be associated.  Once those handles
```
gabrielschulhof: "... newly created handle*s* will be ..." or "... *a* newly created handle will be ..."

napi_memory_management.md
```diff
@@ -0,0 +1,175 @@
...
+scope while it is active.  Scopes must be closed in the reverse order from
+which they are opened.  In addition, all scopes created within a native method
+must be closed before returning from that method.
+
+Taking the earlier example, adding calls to `napi_open_handle_scope()` and
+`napi_close_handle_scope()`  would ensure that at most a single handle
+is valid throughout the execution of the loop:
+
+```
+for (int i = 0; i < 1000000; i++) {napi_
```
gabrielschulhof: "... {~~napi_~~"

napi_memory_management.md
```diff
@@ -0,0 +1,175 @@
...
+active scope at any time, and all new handles will be associated with that
+scope while it is active.  Scopes must be closed in the reverse order from
+which they are opened.  In addition, all scopes created within a native method
+must be closed before returning from that method.
+
+Taking the earlier example, adding calls to `napi_open_handle_scope()` and
+`napi_close_handle_scope()`  would ensure that at most a single handle
+is valid throughout the execution of the loop:
+
+```
```
gabrielschulhof: "&#96;&#96;&#96;*C*"

napi_memory_management.md
```diff
@@ -0,0 +1,175 @@
...
+  if (status != napi_ok) {
+    break;
+  }
+}
+```
+
+When nesting scopes, there are cases where you want a handle from an
+inner scope to live beyond the lifespan of that scope.  N-API supports an
+'escapable scope' in order to support this case.  An escapable scope
+allows one or more handles to be 'promoted' so that they  'escape' the
```
gabrielschulhof: two spaces between "they" and "'escape'"

napi_memory_management.md
```diff
@@ -0,0 +1,175 @@
...
+                                 napi_escapable_handle_scope* result);
+NAPI_EXTERN napi_status
+napi_close_escapable_handle_scope(napi_env env,
+                                  napi_escapable_handle_scope scope);
+```
+
+The request to promote a handle is made through the `napi_escape_handle`
+function:
+
+```
```
gabrielschulhof: "&#96;&#96;&#96;*C*"

napi_memory_management.md
```diff
@@ -0,0 +1,175 @@
...
+NODE_EXTERN napi_status napi_reference_unref(napi_env e,
+                                             napi_ref ref,
+                                             int* result);
+
+NODE_EXTERN napi_status napi_get_reference_value(napi_env e,
+                                                 napi_ref ref,
+                                                 napi_value* result);
+```
+
+the napi_value passed in or out of these methods is a handle to the
```
gabrielschulhof: &#96;napi_value&#96;

gabrielschulhof (approved):
Just a few minor nits, otherwise LGTM.

mhdawson:
fixed comments, going to land now.

mhdawson:
Landed as 9c4304837183d447419191a2951b5058a70c88f6

PR #176, opened 2017-03-21
Minor API cleanup

jasongin:
- Make env parameter name consistent in header
 - Remove napi_get_cb_holder; it is not used anywhere. We can add it back later if it's really needed.

@digitalinfinity had a long list of minor fit and finish issues he encountered when writing documentation. These two are in my judgement the most significant, obvious, and simple fixes that we can make before submitting the PR.

### DIFF
--- a/src/node_api.cc
+++ b/src/node_api.cc
@@ -1411,22 +1411,6 @@ napi_status napi_get_cb_this(napi_env e,
   return napi_ok;
 }
 
-// Holder is a V8 concept.  Is not clear if this can be emulated with other VMs
-// AFAIK Holder should be the owner of the JS function, which should be in the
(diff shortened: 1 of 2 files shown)


jasongin:
https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-commit-linux-abi/60/

boingoing (approved):
Sounds good to me. Will one of the linters complain if the napi_env parameters are named 'env' in the header but 'e' in the source? I prefer the consistency but am not hard-line about it.

jasongin:
No, the linter apparently doesn't care because they weren't consistent between the .h and .cc files before.

digitalinfinity (approved):

PR #181, opened 2017-03-22
Use size_t where appropriate

jasongin:
Based on feedback in the nodejs/node PR, `size_t` should be preferred over `int` generally when referring to array/buffer lengths.

I did not change uses of `int` that refer to `args` array length -- is that right or should it be `size_t` also?

### DIFF
--- a/src/node_api.cc
+++ b/src/node_api.cc
@@ -659,7 +659,7 @@ napi_status napi_define_class(napi_env e,
                               const char* utf8name,
                               napi_callback constructor,
                               void* data,
-                              int property_count,
+                              size_t property_count,
                               const napi_property_descriptor* properties,
                               napi_value* result) {
   NAPI_PREAMBLE(e);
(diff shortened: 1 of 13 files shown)


jasongin:
CI passed: https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-commit-linux-abi/63/

boingoing (approved):
I am on the fence about ```argc``` being ```size_t``` or ```int```. Looks like ```v8::FunctionCallbackInfo::Length``` returns an int and ```JsCallFunction``` takes argument count as ```unsigned short```. Both cases indicate that ```int``` should be large enough. 

However, there really would be no problem for us to use ```size_t``` for arg count in napi functions. In the v8 port, we could return an error if user code tries to call a function with ```argc > MAX_INT``` and ChakraCore port could return an error if user code tries to call a function with ```argc >= 0xFFFF```. For example.

mhdawson (approved):
LGTM

mhdawson:
I'm leaning to size_t for args length mostly to be consistent.

digitalinfinity (approved):
LGTM. The ECMAScript spec doesn't seem to impose any limits on the number of arguments to a function, so I think changing that to size_t is also more future-proof.

jasongin:
LOL at a JS method that takes more than MAX_INT parameters.

Anyway I agree `size_t` makes sense for args because it's unsigned and consistent with other APIs that deal with array lengths.

jasongin:
CI passed after the second commit:
https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-commit-linux-abi/79/

Issue #184, opened 2017-03-22
Remove [=yes|no] from runtime flag


gabrielschulhof:

jasongin:
At the same time, we should also update the error message to mention the runtime flag, as [suggested in this comment](https://github.com/nodejs/node/pull/11975#discussion_r107325207)?

gabrielschulhof:
Definitely.

*  [ ] Remove yes/no
*  [ ] update docs
*  [ ] mention runtime flag in error

PR #186, opened 2017-03-22
all: update --napi-modules flag to not have a yes/no

gabrielschulhof:
This updates the documentation, the error message upon module load
failure, the command line option parsing of the flag, and the way
the N-API addon tests pass the flag to node.

Re https://github.com/nodejs/node/pull/11975#discussion_r107257386
Re https://github.com/nodejs/node/pull/11975#discussion_r107325207
Fixes https://github.com/nodejs/abi-stable-node/issues/184

### DIFF
--- a/doc/api/cli.md
+++ b/doc/api/cli.md
@@ -144,6 +144,13 @@ added: v6.0.0
 
 Silence all process warnings (including deprecations).
 
+### `--napi-modules`
+<!-- YAML
+added: v8.0.0
+-->
+
+Load N-API modules.
+
 ### `--trace-warnings`
 <!-- YAML
 added: v6.0.0
--- a/doc/api/cli.md
+++ b/doc/api/cli.md
@@ -149,7 +149,7 @@ Silence all process warnings (including deprecations).
 added: v8.0.0
 -->
 
-Load N-API modules.
+Load N-API modules (experimental, opt-in by adding this flag).
 
 ### `--trace-warnings`
 <!-- YAML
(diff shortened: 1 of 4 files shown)


doc/api/cli.md
```diff
@@ -144,6 +144,13 @@ added: v6.0.0
 
 Silence all process warnings (including deprecations).
 
+### `--napi-modules`
+<!-- YAML
+added: v8.0.0
+-->
+
+Load N-API modules.
```
jasongin: Add a note about experimental status?

src/node.cc
```diff
@@ -2471,7 +2471,8 @@ void DLOpen(const FunctionCallbackInfo<Value>& args) {
                sizeof(errmsg),
                "The module '%s'"
                "\nwas compiled against the Node.js API. This feature is "
```
jasongin: Should this say "N-API" instead of "Node.js API"?

gabrielschulhof: Perhaps we should be even more verbose and say "... against the Node.js ABI-stable API." If you google "Node.js ABI-stable API", the first link is to nodejs/abi-stable-node, so this might be a good copy/paster for people looking up error messages.

gabrielschulhof: Well, "Node.js N-API" googles to our page too, but only after saying "Did you mean *Node.jsoN-API*?", so, perhaps we should go with the verbose version.

jasongin: "... against the ABI-stable Node.js API (N-API)" ?

mhdawson: I like @jasongin's last suggestion.

mhdawson (approved):
LGTM with change to message as suggeted

jasongin (approved):

boingoing (approved):
👍  LGTM

gabrielschulhof:
https://ci.nodejs.org/view/All/job/node-test-commit-linux-abi/69/

PR #192, opened 2017-03-22
napi: Add string latin1 apis

sampsongao:


src/node_api.h
```diff
@@ -166,6 +170,18 @@ NAPI_EXTERN napi_status napi_get_value_string_length(napi_env env,
                                                      napi_value value,
                                                      int* result);
 
+// Gets the number of BYTES in the LATIN-1 encoded representation of the string.
+NAPI_EXTERN napi_status napi_get_value_string_latin1_length(napi_env env,
```
jasongin: I'm not sure we need this. Since latin1 is a one-byte encoding, the byte length is the same as the character length, so `napi_get_value_string_length` could be used instead.

sampsongao: It seems to me that napi_get_string_utf16_length is calling String::Length(). So is napi_get_string_utf16_length also redundant?

src/node_api.cc
```diff
@@ -1149,6 +1149,25 @@ napi_status napi_create_array_with_length(napi_env e,
   return GET_RETURN_STATUS();
 }
 
+napi_status napi_create_string_latin1(napi_env e,
+                                      const char* s,
```
mhdawson: e and s should be full words, for example e should b env.  There is another PR fixing this up across the napi codebase so we should make sure we keep it up.

src/node_api.h
jasongin: Taylor is removing `napi_get_value_string_utf8_length` and `napi_get_value_string_utf16_length` in #193 . In discussion, we decided it is cleaner to allow the `napi_get_value_string_*` APIs to optionally accept `nullptr` for the buffer and then return the length in bytes. So the same should be done for latin1.

We are keeping `napi_get_value_string_length` though, because the intention of that is to always get the *character* length.

src/node_api.cc
jasongin: That PR has landed now, so you can rebase and make sure the names and behavior are consistent.

src/node_api.cc
```diff
@@ -1618,6 +1637,30 @@ napi_status napi_get_value_string_length(napi_env e,
...
+                                         napi_value value,
+                                         char* buf,
+                                         int bufsize,
+                                         int* result) {
+  NAPI_PREAMBLE(e);
+
+  v8::Local<v8::Value> val = v8impl::V8LocalValueFromJsValue(value);
+  RETURN_STATUS_IF_FALSE(val->IsString(), napi_string_expected);
+
+  int copied = val.As<v8::String>()->WriteOneByte(
```
jasongin: If `buf` is null, just return the length. Refer to the updated implementation of the other `napi_get_value_string_` APIs, after rebasing.

test/parallel/test-os.js
```diff
@@ -181,7 +181,7 @@ if (common.isWindows) {
 } else {
   is.number(pwd.uid);
   is.number(pwd.gid);
-  assert.ok(pwd.shell.includes(path.sep));
+  //assert.ok(pwd.shell.includes(path.sep));
```
jasongin: Revert this change.

src/node_api.cc
```diff
@@ -1149,6 +1149,25 @@ napi_status napi_create_array_with_length(napi_env e,
   return GET_RETURN_STATUS();
 }
 
+napi_status napi_create_string_latin1(napi_env e,
+                                      const char* s,
+                                      int length,
```
boingoing: Replace ```int``` with ```size_t```.

src/node_api.cc
```diff
@@ -1618,6 +1637,30 @@ napi_status napi_get_value_string_length(napi_env e,
   return GET_RETURN_STATUS();
 }
 
+// Copies a JavaScript string into a LATIN-1 string buffer. The result is the
+// number
```
boingoing: Can you clean the spacing of this comment, update it to indicate how it works when buf is NULL, and remove any non-UTF-8 whitespace characters (IIRC, the comments near other string functions had such characters)?

boingoing (changes requested):
Looks good, couple minor comments.

src/node_api.h
```diff
@@ -170,6 +174,18 @@ NAPI_EXTERN napi_status napi_get_value_string_length(napi_env env,
...
 
+// Copies LATIN-1 encoded bytes from a string into a buffer.
+NAPI_EXTERN napi_status napi_get_value_string_latin1(napi_env env,
+                                                     napi_value value,
+                                                     char* buf,
+                                                     size_t bufsize,
+                                                     size_t* result);
+
+// Gets the number of BYTES in the UTF-8 encoded representation of the string.
+NAPI_EXTERN napi_status napi_get_value_string_utf8_length(napi_env env,
```
gabrielschulhof: This should not be here.

jasongin (changes requested):
Sorry for the delay in reviewing.

The code here looks good, but it should have tests in `test/addons-napi/test_string`. (We need to add tests for UTF-16 also; the tests there are only UTF-8.)

test/addons-napi/test_string/test.js
```diff
@@ -6,21 +6,28 @@ const assert = require('assert');
 const test_string = require(`./build/${common.buildType}/test_string`);
 
 const str1 = 'hello world';
```
jasongin, line 8: Can you add a test case with an empty string?

test/addons-napi/test_string/test_string.c
```diff
@@ -1,6 +1,46 @@
...
+
+  char buffer[128];
+  int buffer_size = 128;
+
+  status =
+      napi_get_value_string_latin1(env, args[0], buffer, buffer_size, NULL);
+  if (status != napi_ok) return;
+
+  napi_value output;
+  status = napi_create_string_latin1(env, buffer, -1, &output);
```
jasongin, line 36: I'm surprised that this does not produce compiler errors. (I realize the existing UTF8 test did this also.)
The length parameter is declared as a `size_t` which is always unsigned, but -1 is allowed?

sampsongao: I remember we allow passing -1 as default to let napi figure out the size of string. Do we still support that feature or not? It seems to me that we now have to pass in a valid length for the string.

jasongin: Yeah I know that used to be intentionally supported, but I assumed it wouldn't work anymore after switching to `size_t`.

jasongin: Oh, I think this only works because it is C code, where any implicit conversions between numeric types are allowed. But C++ would be more strict about disallowing the conversion from -1 to `size_t`.

jasongin: I'm not saying you need to make any changes for this right now. I just noted it as something we might want to look at more closely sometime.

mhdawson:
PR landed in main repo, closing this one

Issue #189, opened 2017-03-22
Use snake_case for all local names and parameters


boingoing:
We aren't consistent in naming our locals or parameters with respect to their casing. Node.js, apparently, prefers snake_casing for these.

```c++
  if (staticPropertyCount > 0) {
    std::vector<napi_property_descriptor> staticDescriptors;
    staticDescriptors.reserve(staticPropertyCount);
```
Should become
```c++
  if (static_property_count > 0) {
    std::vector<napi_property_descriptor> static_descriptors;
    static_descriptors.reserve(static_property_count);
```

boingoing:
Fixed by #193

Issue #188, opened 2017-03-22
Use ref/unref terminology instead of addref/release


jasongin:
Consider renaming:
 - `napi_reference_addref()`
 - `napi_reference_release()`

To:
 - `napi_reference_ref()`
 - `napi_reference_unref()`

To be consistent with the "Ref/Unref" terminology used throughout Node.js.

See https://github.com/nodejs/node/pull/11975#discussion_r107326862

boingoing:
Fixed by #193

PR #193, opened 2017-03-23
Convert all locals and parameter names to snake_case

boingoing:
Standardize some parameter names across the N-API surface such as ```napi_env env``` and ```size_t length```. Make all the parameter names in the header match the cc file and expand some parameter names from single characters into meaningful words.
    
Also addresses a couple of minor PR feedback items:
 - Rename ```napi_reference_addref``` and ```napi_reference_release``` to ```napi_reference_ref``` and ```napi_reference_unref``` (respectively)
 - Rename ```Reference::AddRef``` and ```Reference::Release``` to ```Reference::Ref``` and ```Reference::Unref``` (respectively)
 - Remove ```napi_create_boolean```, ```napi_get_true```, and ```napi_get_false``` and replace them with ```napi_get_boolean```
 - Rename ```napi_get_type_of_value``` to ```napi_typeof```
 - Add a ```result_data``` out parameter to ```napi_create_buffer_copy``` which returns the data pointer of the new buffer
 - Change ```napi_get_value_string_utf8``` and ```napi_get_value_string_utf16``` to accept a null output buffer and return the length of the source string in that case via the ```result``` parameter
 - Remove ```napi_get_value_string_utf8_length``` and ```napi_get_value_string_utf16_length``` due to the above
 - Change ```Reference``` ctor and dtor to be protected and added public ```static Reference* Reference::New``` and ```static void Reference::Delete``` methods to make it more clear how the ```Reference``` objects are allocated and cleaned-up
 - Add a type check to ```napi_get_array_length``` which returns an error if the argument is not an array object
 - Change ```napi_create_symbol``` to take a ```napi_value``` instead of a ```const char*``` for the symbol description string. The API now throws if description is not a string but NULL is allowed.

Addresses https://github.com/nodejs/abi-stable-node/issues/189 and https://github.com/nodejs/abi-stable-node/issues/188 and some feedback in https://github.com/nodejs/node/pull/11975.

### DIFF
--- a/src/node_api.cc
+++ b/src/node_api.cc
@@ -1,4 +1,4 @@
-/******************************************************************************
+﻿/******************************************************************************
  * Experimental prototype for demonstrating VM agnostic and ABI stable API
  * for native modules to use instead of using Nan and V8 APIs directly.
  *
@@ -113,22 +113,22 @@ static v8::Local<v8::Function> V8LocalFunctionFromJsValue(napi_value v) {
 
 // Wrapper around v8::Persistent that implements reference counting.
 class Reference {
- public:
+ protected:
   Reference(v8::Isolate* isolate,
             v8::Local<v8::Value> value,
-            int initialRefcount,
-            bool deleteSelf,
-            napi_finalize finalizeCallback = nullptr,
-            void* finalizeData = nullptr,
-            void* finalizeHint = nullptr)
+            int initial_refcount,
+            bool delete_self,
+            napi_finalize finalize_callback = nullptr,
+            void* finalize_data = nullptr,
+            void* finalize_hint = nullptr)
        : _isolate(isolate),
         _persistent(isolate, value),
-        _refcount(initialRefcount),
-        _deleteSelf(deleteSelf),
-        _finalizeCallback(finalizeCallback),
-        _finalizeData(finalizeData),
-        _finalizeHint(finalizeHint) {
-    if (initialRefcount == 0) {
+        _refcount(initial_refcount),
+        _delete_self(delete_self),
+        _finalize_callback(finalize_callback),
+        _finalize_data(finalize_data),
+        _finalize_hint(finalize_hint) {
+    if (initial_refcount == 0) {
       _persistent.SetWeak(
           this, FinalizeCallback, v8::WeakCallbackType::kParameter);
       _persistent.MarkIndependent();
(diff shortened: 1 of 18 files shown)


jasongin:
> Change napi_get_value_string_utf8 and napi_get_value_string_utf16 to accept a null output buffer and return the length of the source string in that case via the result parameter
Remove napi_get_value_string_utf8_length and napi_get_value_string_utf16_length due to the above

The [latin1 string API PR](https://github.com/nodejs/abi-stable-node/pull/192) will need these changes (whichever PR goes second).

src/node_api.h
```diff
@@ -116,21 +117,18 @@ NAPI_EXTERN napi_status napi_create_array_with_length(napi_env env,
...
                                                       napi_value* result);
 NAPI_EXTERN napi_status napi_create_number(napi_env env,
-                                           double val,
+                                           double value,
                                            napi_value* result);
 NAPI_EXTERN napi_status napi_create_string_utf8(napi_env env,
-                                                const char* s,
-                                                size_t length,
+                                                const char* buf,
+                                                size_t bufsize,
```
jasongin: I'm not sure I like the `buf`, `bufsize` parameter names here. I guess the intention was to make them consistent with the parameters in `napi_get_value_string_` APIs? But to me, "buf" implies a buffer that is *to be filled* by the function that is called.

src/node_api.cc
```diff
@@ -113,22 +113,22 @@ static v8::Local<v8::Function> V8LocalFunctionFromJsValue(napi_value v) {
 
 // Wrapper around v8::Persistent that implements reference counting.
 class Reference {
- public:
+ protected:
```
jasongin: Why not `private`? `protected` implies there might be a subclass, but there isn't.

src/node_api.cc
```diff
@@ -1117,112 +1140,125 @@ napi_status napi_get_prototype(napi_env e,
...
-      v8::Number::New(v8impl::V8IsolateFromJsEnv(e), v));
+      v8::Number::New(v8impl::V8IsolateFromJsEnv(env), value));
 
   return GET_RETURN_STATUS();
 }
 
-napi_status napi_create_boolean(napi_env e, bool b, napi_value* result) {
-  NAPI_PREAMBLE(e);
+napi_status napi_get_boolean(napi_env env, bool value, napi_value* result) {
+  NAPI_PREAMBLE(env);
```
jasongin: NAPI_PREAMBLE (TryCatch) is definitely not needed here.

Also in `napi_get_global()`, `napi_get_undefined()`, `napi_get_null()`. Yeah it is not related to the changes you're making, but it just occurred to me and seems like an easy and obvious thing to fix now.

jasongin (changes requested):
This is a lot of good cleanup work. I just had those 3 minor comments.

src/node_api.cc
boingoing: I don't feel strongly about it. Was reasoning that subclassing ```Reference``` might happen at some point. Will change to ```private```.

src/node_api.h
boingoing: Yes, was trying to be consistent but wasn't entirely satisfied with ```buf``` and ```bufsize```, either. ```bufsize``` seems to indicate byte count when it should be length in characters. I'll rename these ```str``` and ```length``` to clear up confusion.

src/node_api.cc
boingoing: Fair enough. I had a similar thought while making these changes but figured it would make more sense to bundle all such NAPI_PREAMBLE cleanups together. These couple are really simple, though. Will add them.

boingoing:
Passing CI run
https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-commit-linux-abi/85/

mhdawson (approved):
LGTM

jasongin (approved):

PR #196, opened 2017-03-24
Change napi_callback to return napi_value instead of void

boingoing:
Change ```napi_callback``` to return ```napi_value``` instead of requiring the callback to manually call ```napi_set_return_value```.

When we invoke the callback, we will check the return value and call ```SetReturnValue``` ourselves. If the callback returns ```NULL```, we don't set the return value in v8 which would have the same effect as previously if the callback didn't call ```napi_set_return_value```. Seems to be a more natural way to handle return values from callbacks. As a consequence, remove ```napi_set_return_value```.

Add a ```napi_value``` to ```napi_property_descriptor``` to support string values which couldn't be passed in the ```utf8name``` parameter or symbols as property names. Class names, however, cannot be symbols so this ```napi_value``` must be a string type in that case.

Remove all of the ```napi_callback_info``` helpers except for ```napi_get_cb_info``` and make all the parameters to ```napi_get_cb_info``` optional except for argc.

Update all the test collateral according to these changes.

### DIFF
--- a/src/node_api.cc
+++ b/src/node_api.cc
@@ -283,7 +283,11 @@ class CallbackWrapperBase : public CallbackWrapper {
         v8::Local<v8::External>::Cast(
             _cbdata->GetInternalField(InternalFieldIndex))->Value());
     v8::Isolate* isolate = _cbinfo.GetIsolate();
-    cb(v8impl::JsEnvFromV8Isolate(isolate), cbinfo_wrapper);
+    napi_value result = cb(v8impl::JsEnvFromV8Isolate(isolate), cbinfo_wrapper);
+
+    if (result != nullptr) {
+      this->SetReturnValue(result);
+    }
 
     if (!TryCatch::LastException().IsEmpty()) {
       isolate->ThrowException(
@@ -577,6 +581,7 @@ const char* error_messages[] = {nullptr,
                                 "Invalid pointer passed as argument",
                                 "An object was expected",
                                 "A string was expected",
+                                "A string or symbol was expected",
                                 "A function was expected",
                                 "A number was expected",
                                 "A boolean was expected",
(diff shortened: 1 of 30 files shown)


boingoing:
@digitalinfinity @jasongin 

Please take a look, I'll push a further commit to fix lint bugs found via CI.

boingoing:
Passing CI:
https://ci.nodejs.org/job/node-test-commit-linux-abi/93/

src/node_api.cc
```diff
@@ -710,7 +724,11 @@ napi_status napi_define_class(napi_env env,
                                     v8impl::FunctionCallbackWrapper::Invoke,
                                     cbdata,
                                     v8::Signature::New(isolate, tpl));
-      t->SetClassName(property_name);
+
+      // Class name has to be an identifier, though property names can be any
+      // property key. Enforce that class name is a string here.
+      RETURN_STATUS_IF_FALSE(property_name->IsString(), napi_string_expected);
+      t->SetClassName(property_name.As<v8::String>());
```
jasongin: If the property name is not a string, I think this should just skip setting the name of the function, not return an error.

jasongin: Actually I don't think `SetClassName()` should be called at all here. The V8 doc for that API says:

> This is used for printing objects created with the function created from the FunctionTemplate as its constructor. 

What this was intended to do was set the function (class method) name. It's odd that `v8::FunctionTemplate` doesn't have a `SetName()` method, while `v8::Function` does.

jasongin: Probably the function will get its name from the property name automatically, so there is no need to set the name separately.

jasongin (changes requested):
Looks good aside from the one issue about class method names. I do think these are definitely improvements to the API surface. And eliminating `napi_set_return_value` will improve performance a little.

But we should consider whether to hold this until after the node PR lands. Does anyone else have thoughts on that?

boingoing:
Sounds right. I was also confused about why we were calling SetClassName here. Seems that is for constructors, not methods.

src/node_api.cc
boingoing: Sounds right. I was also confused about why we were calling ```SetClassName()``` here. Seems that is for constructors, not methods.

jasongin:
> all the parameters to napi_get_cb_info optional except for argc

Why isn't `argc` optional also? Maybe methods which don't expect (or ignore) any parameters shouldn't need to supply that.

boingoing:
This is a little disruptive, yes, waiting for the PR to land is probably fine. I don't want it to get stale, though, since there are so many invasive changes to the test code. Most of the change is in the test code, which is probably not getting as much attention over in the PR.

boingoing:
> Why isn't argc optional also? Maybe methods which don't expect (or ignore) any parameters shouldn't need to supply that.

I thought about it but figured it didn't make sense because argc is both an in and an out parameter so it's required if argv is provided. That might be fine, though, to say if argv is provided that argc becomes non-optional.

src/node_api.cc
```diff
@@ -282,7 +282,11 @@ class CallbackWrapperBase : public CallbackWrapper {
         v8::Local<v8::External>::Cast(
             _cbdata->GetInternalField(InternalFieldIndex))->Value());
     v8::Isolate* isolate = _cbinfo.GetIsolate();
-    cb(v8impl::JsEnvFromV8Isolate(isolate), cbinfo_wrapper);
+    napi_value result = cb(v8impl::JsEnvFromV8Isolate(isolate), cbinfo_wrapper);
+
+    if (result) {
```
jasongin: Anna just commented on the main PR that node style prefers `if (value != nullptr)`. Also below.

jasongin:
> argv is provided that argc becomes non-optional

Yes. While updating leveldown code I encountered several places where it would be more convenient to pass nullptr for both argc and argv.

src/node_api.cc
boingoing: Saw that and updated uses in this PR.

boingoing:
Another CI job after rebasing
https://ci.nodejs.org/job/node-test-commit-linux-abi/96/

mhdawson:
landed in main repo closing.

Issue #124, opened 2017-03-02
NAPI number coercion behavior


jasongin:
The `node_sass` package calls `Nan::To<int32_t>(value)`, which calls `value->Int32Value()`, that returns `0` if the value is undefined.

In the NAPI V8 implementation, `napi_coerce_to_number()` calls `value->ToNumber()`, and `napi_get_value_int32()` calls `value->Int32Value()`, which returns `INT_MIN` if the original value before coercion was undefined.

So while it seemed obvious when porting to convert `Nan::To<int32_t>()` to `napi_coerce_to_number()` and `napi_get_value_int32()`, the result was unexpected.

Do we need a `napi_coerce_to_int32()` (and similar for other primitive types) that would provide the behavior that was expected here?

jasongin:
I verified using an updated `node_sass` that this issue was fixed by https://github.com/nodejs/abi-stable-node/pull/199.

PR #199, opened 2017-03-24
Updates for review feedback

jasongin:
- Delete the `node_api_*internal.h` headers - there is no longer any reason for their contents to be separate from the `.cc` files.
 - Always set callback data internal field to prevent crashing when `V8_ENABLE_CHECKS` is defined
 - Don't crash when casting numbers when `V8_ENABLE_CHECKS` is defined
 - Return 0 when getting NaN value as an integer (also fixes https://github.com/nodejs/abi-stable-node/issues/124)
 - Style: replace `if (ptr)` with `if (ptr != nullptr)`
 - Fix extern "C" in `node_api_async.h`
 - Minor doc changes


jasongin:
Addresses all of today's feedback on the PR at https://github.com/nodejs/node/pull/11975

jasongin:
CI passed: https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-commit-linux-abi/95/

digitalinfinity:
:shipit:

digitalinfinity (approved):

boingoing (approved):
LGTM

PR #204, opened 2017-03-28
Redesign async helper APIs

boingoing:
Add error handling for the async helper APIs. At least they all return ```napi_status``` now.

Remove the individual APIs for setting all the pieces of the ```napi_work``` struct opting to just take all the parameters in ```napi_create_async_work```. It doesn't seem like any of the parameters are optional here anyway.
    
Renamed the ```napi_work``` struct to ```napi_async_work``` and replace the struct itself with a class which can better encapsulate the object lifetime and ```uv_work_t``` that we're trying to wrap anyway.
    
Added a ```napi_async_callback``` type for the async helper callbacks instead of taking raw function pointers and make this callback take a ```napi_env``` parameter as well as the ```void* data``` it was already taking.
    
Also added some async unit tests for addons-napi based on the addons/async_hello_world test.

### DIFF
--- a/src/node_api.cc
+++ b/src/node_api.cc
@@ -16,6 +16,7 @@
 #include <cmath>
 #include <vector>
 #include "node_api.h"
+#include "env.h"
 
 namespace v8impl {
 
@@ -627,7 +628,8 @@ const char* error_messages[] = {nullptr,
                                 "A boolean was expected",
                                 "An array was expected",
                                 "Unknown failure",
-                                "An exception is pending"};
+                                "An exception is pending",
+                                "The async work item was cancelled"};
 
 void napi_clear_last_error() {
   static_last_error.error_code = napi_ok;
@@ -2527,4 +2529,137 @@ napi_status napi_get_typedarray_info(napi_env env,
   }
 
   return GET_RETURN_STATUS();
+}
+
+namespace uvimpl {
+
+napi_status ConvertUVErrorCode(int code) {
+  switch (code) {
+  case 0:
+    return napi_ok;
+  case UV_EINVAL:
+    return napi_invalid_arg;
+  case UV_ECANCELED:
+    return napi_cancelled;
+  }
+
+  return napi_generic_failure;
+}
+
+// Wrapper around uv_work_t which calls user-provided callbacks.
+class Work {
+ private:
+  explicit Work(napi_env env,
+                napi_async_execute_callback execute = nullptr,
+                napi_async_complete_callback complete = nullptr,
+                void* data = nullptr)
+    : _env(env),
+    _data(data),
+    _execute(execute),
+    _complete(complete) {
+    _request.data = this;
+  }
+
+  ~Work() { }
+
+ public:
+  static Work* New(napi_env env,
+                   napi_async_execute_callback execute,
+                   napi_async_complete_callback complete,
+                   void* data) {
+    return new Work(env, execute, complete, data);
+  }
+
+  static void Delete(Work* work) {
+    delete work;
+  }
+
+  static void ExecuteCallback(uv_work_t* req) {
+    Work* work = static_cast<Work*>(req->data);
+    work->_execute(work->_env, work->_data);
+  }
+
+  static void CompleteCallback(uv_work_t* req, int status) {
+    Work* work = static_cast<Work*>(req->data);
+
+    if (work->_complete != nullptr) {
+      work->_complete(work->_env, ConvertUVErrorCode(status), work->_data);
+    }
+  }
+
+  uv_work_t* Request() {
+    return &_request;
+  }
+
+ private:
+  napi_env _env;
+  void* _data;
+  uv_work_t _request;
+  napi_async_execute_callback _execute;
+  napi_async_complete_callback _complete;
+};
+
+}  // end of namespace uvimpl
+
+#define CALL_UV(condition)                                              \
+  do {                                                                  \
+    int result = (condition);                                           \
+    napi_status status = uvimpl::ConvertUVErrorCode(result);            \
+    if (status != napi_ok) {                                            \
+      return napi_set_last_error(status, result);                       \
+    }                                                                   \
+  } while (0)
+
+napi_status napi_create_async_work(napi_env env,
+                                   napi_async_execute_callback execute,
+                                   napi_async_complete_callback complete,
+                                   void* data,
+                                   napi_async_work* result) {
+  CHECK_ARG(env);
+  CHECK_ARG(execute);
+  CHECK_ARG(result);
+
+  uvimpl::Work* work = uvimpl::Work::New(env, execute, complete, data);
+
+  *result = reinterpret_cast<napi_async_work>(work);
+
+  return napi_ok;
+}
+
+napi_status napi_delete_async_work(napi_env env, napi_async_work work) {
+  CHECK_ARG(env);
+  CHECK_ARG(work);
+
+  uvimpl::Work::Delete(reinterpret_cast<uvimpl::Work*>(work));
+
+  return napi_ok;
+}
+
+napi_status napi_queue_async_work(napi_env env, napi_async_work work) {
+  CHECK_ARG(env);
+  CHECK_ARG(work);
+
+  // Consider: Encapsulate the uv_loop_t into an opaque pointer parameter
+  uv_loop_t* event_loop = node::Environment::GetCurrent(
+    v8impl::V8IsolateFromJsEnv(env))->event_loop();
+
+  uvimpl::Work* w = reinterpret_cast<uvimpl::Work*>(work);
+
+  CALL_UV(uv_queue_work(event_loop,
+                        w->Request(),
+                        uvimpl::Work::ExecuteCallback,
+                        uvimpl::Work::CompleteCallback));
+
+  return napi_ok;
+}
+
+napi_status napi_cancel_async_work(napi_env env, napi_async_work work) {
+  CHECK_ARG(env);
+  CHECK_ARG(work);
+
+  uvimpl::Work* w = reinterpret_cast<uvimpl::Work*>(work);
+
+  CALL_UV(uv_cancel(reinterpret_cast<uv_req_t*>(w->Request())));
+
+  return napi_ok;
 }
--- a/test/addons-napi/test_async/test.js
+++ b/test/addons-napi/test_async/test.js
@@ -1,0 +1,10 @@
+'use strict';
+const common = require('../../common');
+const assert = require('assert');
+const test_async = require(`./build/${common.buildType}/test_async`);
+
+test_async(5, common.mustCall(function(err, val) {
+  assert.strictEqual(err, null);
+  assert.strictEqual(val, 10);
+  process.nextTick(common.mustCall(function() {}));
+}));
--- a/src/node_api.h
+++ b/src/node_api.h
@@ -474,6 +474,21 @@ NAPI_EXTERN napi_status napi_get_typedarray_info(napi_env env,
                                                  void** data,
                                                  napi_value* arraybuffer,
                                                  size_t* byte_offset);
+
+// Methods to manage simple async operations
+NAPI_EXTERN
+napi_status napi_create_async_work(napi_env env,
+                                   napi_async_execute_callback execute,
+                                   napi_async_complete_callback complete,
+                                   void* data,
+                                   napi_async_work* result);
+NAPI_EXTERN napi_status napi_delete_async_work(napi_env env,
+                                               napi_async_work work);
+NAPI_EXTERN napi_status napi_queue_async_work(napi_env env,
+                                              napi_async_work work);
+NAPI_EXTERN napi_status napi_cancel_async_work(napi_env env,
+                                               napi_async_work work);
+
 EXTERN_C_END
 
 #endif  // SRC_NODE_API_H__
--- a/src/node_api_types.h
+++ b/src/node_api_types.h
@@ -16,12 +16,7 @@ typedef struct napi_ref__ *napi_ref;
 typedef struct napi_handle_scope__ *napi_handle_scope;
 typedef struct napi_escapable_handle_scope__ *napi_escapable_handle_scope;
 typedef struct napi_callback_info__ *napi_callback_info;
-
-typedef void (*napi_callback)(napi_env env,
-                              napi_callback_info info);
-typedef void (*napi_finalize)(napi_env env,
-                              void* finalize_data,
-                              void* finalize_hint);
+typedef struct napi_async_work__ *napi_async_work;
 
 typedef enum {
   napi_default = 0,
@@ -33,18 +28,6 @@ typedef enum {
   // from instance properties. Ignored by napi_define_properties.
   napi_static_property = 1 << 10,
 } napi_property_attributes;
-
-typedef struct {
-  const char* utf8name;
-
-  napi_callback method;
-  napi_callback getter;
-  napi_callback setter;
-  napi_value value;
-
-  napi_property_attributes attributes;
-  void* data;
-} napi_property_descriptor;
 
 typedef enum {
   // ES6 types (corresponds to typeof)
@@ -82,8 +65,32 @@ typedef enum {
   napi_array_expected,
   napi_generic_failure,
   napi_pending_exception,
+  napi_cancelled,
   napi_status_last
 } napi_status;
+
+typedef void (*napi_callback)(napi_env env, 
+                              napi_callback_info info);
+typedef void (*napi_finalize)(napi_env env,
+                              void* finalize_data, 
+                              void* finalize_hint);
+typedef void (*napi_async_execute_callback)(napi_env env, 
+                                            void* data);
+typedef void (*napi_async_complete_callback)(napi_env env,
+                                             napi_status status,
+                                             void* data);
+
+typedef struct {
+  const char* utf8name;
+
+  napi_callback method;
+  napi_callback getter;
+  napi_callback setter;
+  napi_value value;
+
+  napi_property_attributes attributes;
+  void* data;
+} napi_property_descriptor;
 
 typedef struct {
   const char* error_message;
--- a/test/addons-napi/test_async/binding.gyp
+++ b/test/addons-napi/test_async/binding.gyp
@@ -1,0 +1,8 @@
+{
+  "targets": [
+    {
+      "target_name": "test_async",
+      "sources": [ "test_async.cc" ]
+    }
+  ]
+}
--- a/test/addons-napi/test_async/test_async.cc
+++ b/test/addons-napi/test_async/test_async.cc
@@ -1,0 +1,130 @@
+#include <node_api.h>
+
+#if defined _WIN32
+#include <windows.h>
+#else
+#include <unistd.h>
+#endif
+
+#define NAPI_CHECK(env, condition, msg)                                   \
+  if (!(condition)) {                                                     \
+    napi_throw_error((env), (msg));                                       \
+    return;                                                               \
+  }
+
+#define NAPI_CALL(env, theCall)                                           \
+  if ((theCall) != napi_ok) {                                             \
+    const napi_extended_error_info* error;                                \
+    napi_get_last_error_info((env), &error);                              \
+    const char *errorMessage = error->error_message;                      \
+    errorMessage = errorMessage ? errorMessage : "empty error message";   \
+    napi_throw_error((env), errorMessage);                                \
+    return;                                                               \
+  }
+
+typedef struct {
+  int32_t _input;
+  int32_t _output;
+  napi_ref _callback;
+  napi_async_work _request;
+} carrier;
+
+carrier theCarrier;
+
+struct AutoHandleScope {
+  AutoHandleScope(napi_env env)
+  : _env(env),
+  _scope(NULL) {
+    napi_open_handle_scope(_env, &_scope);
+  }
+  ~AutoHandleScope() {
+    napi_close_handle_scope(_env, _scope);
+  }
+private:
+  AutoHandleScope() { }
+
+  napi_env _env;
+  napi_handle_scope _scope;
+};
+
+void Execute(napi_env env, void* data) {
+#if defined _WIN32
+  Sleep(1000);
+#else
+  sleep(1);
+#endif
+  carrier* c = static_cast<carrier*>(data);
+  
+  if (c != &theCarrier) {
+    napi_throw_type_error(env, "Wrong data parameter to Execute.");
+    return;
+  }
+  
+  c->_output = c->_input * 2;
+}
+
+void Complete(napi_env env, napi_status status, void* data) {
+  AutoHandleScope scope(env);
+  carrier* c = static_cast<carrier*>(data);
+
+  if (c != &theCarrier) {
+    napi_throw_type_error(env, "Wrong data parameter to Complete.");
+    return;
+  }
+  
+  if (status != napi_ok) {
+    napi_throw_type_error(env, "Execute callback failed.");
+    return;
+  }
+
+  napi_value argv[2];
+  
+  NAPI_CALL(env, napi_get_null(env, &argv[0]));
+  NAPI_CALL(env, napi_create_number(env, c->_output, &argv[1]));
+  napi_value callback;
+  NAPI_CALL(env, napi_get_reference_value(env, c->_callback, &callback));
+  napi_value global;
+  NAPI_CALL(env, napi_get_global(env, &global));
+
+  napi_value result;
+  NAPI_CALL(env, napi_call_function(env, global, callback, 2, argv, &result));
+  
+  NAPI_CALL(env, napi_delete_reference(env, c->_callback));
+  NAPI_CALL(env, napi_delete_async_work(env, c->_request));
+}
+
+void Test(napi_env env, napi_callback_info info) {
+  size_t argc = 2;
+  napi_value argv[2];
+  napi_value _this;
+  void* data;
+  NAPI_CALL(env,
+    napi_get_cb_info(env, info, &argc, argv, &_this, &data));
+  NAPI_CHECK(env, argc >= 2, "Not enough arguments, expected 2.");
+
+  napi_valuetype t;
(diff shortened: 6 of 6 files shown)


src/node_api_async.cc
```diff
@@ -1,69 +1,105 @@
...
+};
 }
 
-void napi_async_execute(uv_work_t* req) {
-  napi_work_impl* worker = static_cast<napi_work_impl*>(req->data);
-  worker->execute(worker->data);
+napi_status napi_create_async_work(napi_env env,
+                                   napi_async_callback execute,
+                                   napi_async_callback complete,
+                                   napi_async_callback destroy,
```
sampsongao: I think we can leave destroy callback as optional, because nanomsg do clean up in complete and never have a destroy callback.

src/node_api_async.h
```diff
@@ -7,13 +7,16 @@
...
-NAPI_EXTERN void napi_async_queue_worker(napi_work w);
+NAPI_EXTERN napi_status napi_create_async_work(napi_env env,
+                                               napi_async_callback execute,
+                                               napi_async_callback complete,
+                                               napi_async_callback destroy,
+                                               void* data,
+                                               napi_async_work* result);
+NAPI_EXTERN napi_status napi_delete_async_work(napi_env env, 
+                                               napi_async_work work);
+NAPI_EXTERN napi_status napi_queue_async_work(napi_env env,
```
jasongin: Should there be a `napi_cancel_async_work()` API that calls [`uv_cancel()`](http://docs.libuv.org/en/v1.x/request.html#c.uv_cancel) ?

src/node_api_async.cc
```diff
@@ -1,69 +1,105 @@
...
+
+  static void CompleteCallback(uv_work_t* req, int status) {
+    // Consider: Pass the status into the complete callback?
+    Work* work = reinterpret_cast<Work*>(req->data);
+	
+	if (work->_complete != nullptr) {
+      work->_complete(work->_env, work->_data);
+	}
+	if (work->_destroy != nullptr) {
+      work->_destroy(work->_env, work->_data);
```
jasongin: Why is the destroy callback called here instead of from the `Delete()` method? What is the purpose of having two callbacks (complete and destroy) that are invoked at the same time with the same condition?

src/node_api_async.cc
```diff
@@ -1,69 +1,105 @@
...
+
+  static void ExecuteCallback(uv_work_t* req) {
+    Work* work = reinterpret_cast<Work*>(req->data);
+	
+	if (work->_execute != nullptr) {
+      work->_execute(work->_env, work->_data);
+	}
+  }
+
+  static void CompleteCallback(uv_work_t* req, int status) {
```
jasongin: The status parameter here is ignored. Should it be passed to the complete callback?

src/node_api_async.cc
```diff
@@ -1,69 +1,105 @@
...
+    if (result != 0) {
+      return napi_generic_failure;
+    }
+    return napi_ok;
+  }
+
+  static void ExecuteCallback(uv_work_t* req) {
+    Work* work = reinterpret_cast<Work*>(req->data);
+	
+	if (work->_execute != nullptr) {
```
jasongin: Does it even make sense to queue async work with a null execute callback? Or should `napi_create_async_work()` return an error in that case?

src/node_api_async.cc
```diff
@@ -1,69 +1,105 @@
...
+                               work->_request,
+                               ExecuteCallback,
+                               CompleteCallback);
 
-void napi_async_set_complete(napi_work w, void (*complete)(void* data)) {
-  napi_work_impl* worker = reinterpret_cast<napi_work_impl*>(w);
-  worker->complete = complete;
-}
+    if (result != 0) {
+      return napi_generic_failure;
```
jasongin: Maybe add a helper method to do basic translation between common libuv error codes and NAPI status codes. For example `UV_EINVAL` => `napi_invalid_arg`.

src/node_api_async.cc
```diff
@@ -1,69 +1,105 @@
...
+  static void ExecuteCallback(uv_work_t* req) {
+    Work* work = reinterpret_cast<Work*>(req->data);
+	
+	if (work->_execute != nullptr) {
+      work->_execute(work->_env, work->_data);
+	}
+  }
+
+  static void CompleteCallback(uv_work_t* req, int status) {
+    // Consider: Pass the status into the complete callback?
```
jasongin: Yes, I think it should. Especially if cancelling a work item is supported.

src/node_api_async.cc
```diff
@@ -1,69 +1,105 @@
...
+    delete work;
+  }
 
-void napi_async_set_execute(napi_work w, void (*execute)(void* data)) {
-  napi_work_impl* worker = reinterpret_cast<napi_work_impl*>(w);
-  worker->execute = execute;
-}
+  static napi_status Queue(Work* work) {
+    // Consider: Choose a different loop based on the napi_env?
+    int result = uv_queue_work(uv_default_loop(),
```
jasongin: There isn't necessarily a relationship between the `napi_env` (`v8::Isolate`) and libuv loops, or if there is it doesn't need to be followed by native modules.

In order to support a non-default loop, N-API would have to either wrap most of [the libuv loop APIs](http://docs.libuv.org/en/v1.x/loop.html), or allow an optional `uv_loop_t` parameter to `napi_create_async_work()` (breaking the abstraction barrier). I don't really like either of those alternatives right now.

Another consideration is that the libuv docs [say](http://docs.libuv.org/en/v1.x/loop.html#c.uv_default_loop):

> the default loop can (and should) be closed with uv_loop_close() so the resources associated with it are freed

and N-API currently doesn't offer any way to do that.

gabrielschulhof:
```
$ git grep -nH uv_loop_new
deps/uv/docs/src/migration_010_100.rst:15:In libuv 0.10 (and previous versions), loops were created with `uv_loop_new`, which
deps/uv/docs/src/migration_010_100.rst:24:    uv_loop_t* loop = uv_loop_new();
deps/uv/include/uv.h:267:UV_EXTERN uv_loop_t* uv_loop_new(void);
deps/uv/src/uv-common.c:607:uv_loop_t* uv_loop_new(void) {
```
Looks like node itself only ever uses the default loop, so I don't think we should support multiple loops in any way.

jasongin: Of course node just uses the default loop because all the JavaScript just runs on the one loop thread. But that certainly doesn't mean native modules only use that loop. Native modules that do background processing are very likely to use a non-default loop.

jasongin: Also, if/when node ever supports multiple isolates, I'd expect each to use a separate loop. Still, there's no reason a native module needs to share the event loop with node.

jasongin: See also the comment about this on the PR to node: https://github.com/nodejs/node/pull/11975#discussion_r108029533

src/node_api_async.h
boingoing: Probably. I didn't add any new APIs in this change since I just wanted to get what was already there cleaned-up a little.

src/node_api_async.cc
boingoing: The existing behavior was to have a separate callback for complete and destroy. I guess it might make sense if we were using something other than uv which supported these different cleanup callbacks natively. Agreed, though, it does seem a little weird and I considered removing the destroy callback and expecting user code to clean up the ```napi_async_work``` in the complete handler. I don't think this should call ```Delete()``` directly, by the way, unless we want the ```Work``` class to take ownership of it's own lifetime which we aren't doing now because we hand back the ```napi_async_work``` pointer to user code.

src/node_api_async.cc
boingoing: Yes, there are a bunch of uv error constants so I didn't want to write an exhaustive conversion function but looking at uv_queue_work, yeah, it's only going to return ```UV_EINVAL``` on failure. Will add.

src/node_api_async.cc
boingoing: Hmm. Makes sense. Will fix up ```napi_create_async_work``` to check for null execute callback.

src/node_api_async.cc
boingoing: Should we try to translate it into a ```napi_status```?

src/node_api_async.cc
jasongin: Sure, there's no need for an exhaustive mapping now. But if there is a shared helper function then it will be easy to add additional mappings later when we add new NAPI status codes and/or wrap additional libuv APIs.

src/node_api_async.cc
boingoing: Should we try to abstract this in N-API? We could use an opaque pointer named ```napi_async_execution_context``` which could hold onto the ```uv_loop_t*```. Manage them via ```napi_create_async_execution_context``` and ```napi_destroy_async_execution_context```...

src/node_api_async.cc
boingoing: Should we just remove the destroy callback entirely?

jasongin:
@boingoing also pointed out offline that these APIs need to add support for `napi_get_last_error_info()`.

In my opinion, we should consider removing these APIs from the node PR for now, instead of rushing to get all these changes ready in time. Then we can take the time to get the design right later, potentially as part of a broader treatment for libuv APIs. (If we still think the abstraction is needed at all for libuv at all.)

src/node_api_async.cc
mhdawson: As with other API's we did inherit the basic structure from either the libuv or NaN APIs.  We should tighten up where it makes sense.

src/node_api_async.cc
mhdawson: Given that I don't think any of the add-ons we've ported so far have used an alternative loop, I'd limit ourselves to what is needed/used by the AsyncWorker class in the C++ wrapper.  

We should think it through so we are comfortable that we can add methods to support alternate loops but I don't think we need to necessarily implement them right now.  

I'd also say that we should not assume napi_env maps directly to v8::Isolate as we should be careful to retain enough flexibility to bundle into napi_env whatever is necessary for N-API regardless of the underlying impl.

mhdawson:
I see lots of good improvements in this PR/discussion and I think we should continue to refine and then see what it looks like.  I don't think we should rush to remove, Anna reacted positively to the explanation for why they were there. I really think we should avoid the case were we lead module maintainers down the path of in-inadvertently pulling in methods which are not ABI safe.  If exclude these I think we'll need the C++ wrapper to clearly state up front that it uses non-abi stable methods and as such does not preserve the ABI stability goal.  I don't think that would be a good starting point for the wrapper or N-API.

jasongin:
> If exclude these I think we'll need the C++ wrapper to clearly state up front that it uses non-abi stable methods and as such does not preserve the ABI stability goal.

But libuv APIs *are* ABI-stable. [Its last breaking change was pre-1.0.](http://docs.libuv.org/en/v1.x/migration_010_100.html) While hypothetically it could break compatibility with a 2.x release, no 2.x is planned as far as I know, and may never happen.

Also, if this specific issue is a concern, we could consider removing the `AsyncWorker` class from the C++ wrapper library. Its implementation is a fairly trivial composition of other N-API wrappers (`Napi::ObjectReference` and `Napi::FunctionReference`) with the libuv async work callbacks.

src/node_api_async.cc
jasongin: @gabrielschulhof Node actually does use non-default uv loops in some places. [`uv_loop_new()` is an obsolete API](http://docs.libuv.org/en/v1.x/migration_010_100.html#loop-initialization-and-closing); look for `uv_loop_init()` instead:

```
deps\uv\include\uv.h:260:UV_EXTERN int uv_loop_init(uv_loop_t* loop);
deps\uv\include\uv.h:265: *  allocate the loop manually and use uv_loop_init instead.
src\debug-agent.cc:77:  err = uv_loop_init(&child_loop_);
src\inspector_agent.cc:547:  int err = uv_loop_init(&child_loop_);
src\node_watchdog.cc:36:  rc = uv_loop_init(loop_);
src\spawn_sync.cc:466:  CHECK_EQ(uv_loop_init(uv_loop_), 0);
src\tracing\agent.cc:19:  int err = uv_loop_init(&tracing_loop_);
test\cctest\test_inspector_socket.cc:364:    GTEST_ASSERT_EQ(0, uv_loop_init(&loop));
test\cctest\test_inspector_socket_server.cc:74:    EXPECT_EQ(0, uv_loop_init(&loop));
```

addaleax: Would just using `node::Environment::GetCurrent(isolate)->event_loop()` be okay here?

addaleax:
> While hypothetically it could break compatibility with a 2.x release, no 2.x is planned as far as I know, and may never happen.

A 2.x will happen at *some*, afaik yet indeterminate, point in the future. No changes to the APIs relevant for async work are currently in libuv/libuv@master and no open PRs with related changes exist, though.

src/node_api_async.cc
jasongin: That seems better than just using the default loop. But I think some native modules will need the flexibility to use a separate loop for background processing.

addaleax: Those native modules would use the libuv APIs directly anyway, though?

addaleax: > look for `uv_loop_init()` instead:

Fwiw, none of the other loops should currently be accessible to userland or native addons, I think. When I wrote the corresponding comment in the nodejs/node PR, I had embedders of Node mind, which might spin up multiple Isolates/`node::Environment`s.

jasongin: > Those native modules would use the libuv APIs directly anyway, though?

If they do, then what is the purpose of abstracting those APIs here?

addaleax: > If they do, then what is the purpose of abstracting those APIs here?

Supporting native modules that don’t have their own loop and want a *simple* way to queue async work, I think?

boingoing:
Pushed a commit addressing some review feedback and iterating on the design a little bit. Sorry in advance, this moves big blocks of code and deletes a couple of files so I realize reviewing the diff will be trickier.

Here's the commit message:

> Remove node_api_async files
> Put the async N-API methods into node_api.h/cc so they can access the error handling helpers without having to rearrange that code. I'm not committed to having all our methods in one set of files but this is a relatively small set of things to have an entire parallel set of files to support.
> 
> Add ```napi_cancel_async_work``` for cancelling a ```napi_async_work```.
> 
> Call the complete handler for the async work item with a ```napi_status``` code translated from the uvlib error code.
> 
> Made the execute callback required for ```napi_create_async_work```, though complete callback is still optional.
> 
> Also removed the destroy callback as there doesn't seem to be a reason to support it.

src/node_api.cc
```diff
@@ -2461,3 +2463,139 @@ napi_status napi_get_typedarray_info(napi_env env,
...
+ private:
+  explicit Work(napi_env env,
+                napi_async_execute_callback execute = nullptr,
+                napi_async_complete_callback complete = nullptr,
+                void* data = nullptr)
+    : _env(env),
+    _data(data),
+    _execute(execute),
+    _complete(complete) {
+    _request = new uv_work_t;
```
addaleax: Why not just turn `_request` into a plain `uv_work_t` member, no pointers or so?

src/node_api.cc
```diff
@@ -2461,3 +2463,139 @@ napi_status napi_get_typedarray_info(napi_env env,
...
+                   void* data) {
+    return new Work(env, execute, complete, data);
+  }
+
+  static void Delete(Work* work) {
+    delete work;
+  }
+
+  static void ExecuteCallback(uv_work_t* req) {
+    Work* work = reinterpret_cast<Work*>(req->data);
```
addaleax: static_cast?

Also, if you turn `_request` into a plain data member, you could forget about `req->data` and use Node’s evil `ContainerOf` magic, which is nice because it avoids an unnecessary pointer lookup. (Not that performance really matters here or anything :smile:)

test/addons-napi/test_async/test_async.c
```diff
@@ -0,0 +1,98 @@
...
+    errorMessage = errorMessage ? errorMessage : "empty error message";   \
+    napi_throw_error((env), errorMessage);                                \
+    return (retval);                                                      \
+  }
+
+#define NAPI_CALL(env, theCall) NAPI_CALL_RETURN(env, theCall, void)
+
+typedef struct {
+  int32 _input;
+  int32 _output;
```
addaleax: huh … should this be `int32_t`?

test/addons-napi/test_async/test_async.c
```diff
@@ -0,0 +1,98 @@
...
+
+carrier theCarrier;
+
+void Execute(napi_env env, void* data) {
+#if defined _WIN32
+  Sleep(1000);
+#else
+  sleep(1);
+#endif
+  struct carrier* c = reinterpret_cast<struct carrier*>(data);
```
addaleax: This is a C file, though? :smile:

test/addons-napi/test_async/test.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+const common = require('../../common');
+const assert = require('assert');
+const test_array = require(`./build/${common.buildType}/test_array`);
+
+test_array(5, common.mustCall(function(err, val) {
+  assert.strictEqual(err, null);
+  assert.strictEqual(val, 10);
+  process.nextTick(common.mustCall(function() {}));
```
addaleax: This line is a bit odd… does it serve some particular purpose? It brings up a good point – ideally, the C callback should use `napi_make_callback` instead of `napi_call_function`…

It might even be better if the `complete` callback would always be called in a MakeCallback-ish way, but that probably requires some refactoring of `MakeCallback` itself.

test/addons-napi/test_async/test_async.c
```diff
@@ -0,0 +1,98 @@
...
+  NAPI_CALL(napi_get_null(env, &argv[0]));
+  NAPI_CALL(napi_create_number(env, c->_output, &argv[1]));
+  napi_value callback;
+  NAPI_CALL(napi_get_reference_value(env, c->_callback, &callback));
+  napi_value global;
+  NAPI_CALL(napi_get_global(env, &global));
+
+  NAPI_CALL(napi_call_function(env, global, callback, 2, argv, NULL));
+  
+  struct carrier* c = reinterpret_cast<struct carrier*>(data);
```
addaleax: Maybe add a check to make sure that `c == &theCarrier`? (ditto for Execute)

addaleax (approved):
Looks good, commented with some nice-to-haves

test/addons-napi/test_async/test.js
boingoing: I adapted this test from test/addons/async-hello-world which has this pattern. I guess they're using the empty function here to make sure that the complete callback is actually called? Good point about ```MakeCallback```. I wonder how often the ```complete``` callback would be a c++ function vs some wrapper around JavaScript?

src/node_api.cc
boingoing: This is a real good point.

node.gyp
```diff
@@ -166,9 +166,6 @@
         'src/node_api.cc',
         'src/node_api.h',
         'src/node_api_types.h',
-        'src/node_api_async.cc',
-        'src/node_api_async.h',
-        'src/node_api_async_types.h',
```
jasongin: Also remove the header files from `tools/install.py`

Makefile
```diff
@@ -277,7 +277,6 @@ test/addons-napi/.buildstamp: config.gypi \
 	$(ADDONS_NAPI_BINDING_GYPS) $(ADDONS_NAPI_BINDING_SOURCES) \
 	deps/uv/include/*.h deps/v8/include/*.h \
 	src/node.h src/node_buffer.h src/node_object_wrap.h src/node_version.h \
-	src/node_api.h src/node_api_async.h
```
jasongin: Careful, we need to keep `node_api.h` here.

jasongin: And why isn't `node_api_types.h` listed here?

boingoing: Ah, great catch here, thanks. 👍

test/addons-napi/test_async/test_async.c
boingoing: Good call, I'll add these.

src/node_api.cc
boingoing: I looked into it briefly but the machine I'm on right now has a broken instance of VS, apparently, so I can't compile and test this one. Will see if I can make it work in the morning.

test/addons-napi/test_async/test_async.c
boingoing: uh... yeah.

kkoopa:
> Also, if this specific issue is a concern, we could consider removing the AsyncWorker class from the C++ wrapper library. Its implementation is a fairly trivial composition of other N-API wrappers (Napi::ObjectReference and Napi::FunctionReference) with the libuv async work callbacks.

No, no. It is nowhere near a trivial composition for the average person developing a native addon. The primary reason for writing a native addon is necessity. Many addons are cobbled together from looking at existing addons, without much understanding of what goes on or what errors there are. These then trickle up the dependency chain and break the ecosystem. There is a definite need of cookie-cutter functionality for common use cases, which ensures that things are done right and reduce boilerplate.

Besides ABI stability, I do not consider it good that writing a native addon correctly requires intimate knowledge of the node.js API, the libuv API as well as how they interact. There should be one node.js API with one set of documentation and sensible names that fit together and make a unified whole. Writing an asynchronous native node.js addon should not even require knowing that libuv exists or what it is.

jasongin:
@kkoopa The suggestion about removing AsyncWorker was only meant to be _temporary_, probably only a few days up to a week or two. Sorry if that wasn't clear. We know that the coverage of N-API is not yet complete anyway, which is why it going in as "experimental" status.

Another way to look at it is just breaking the one big PR into two smaller ones that would land separately, to reduce the scope and complexity of each piece.

aruneshchandra:
We had a quick core team sync-up on this issue this morning - here's the takeaway.

There are known gaps in the experimental N-API version today. These gaps are both V8 and libuv related, and we know we have to address them as this project moves along and exits experimental status.  _**We all agree that, at least in the short term, adding ABI stable layer over NAN supported libuv APIs is important for N-API promise and to provide NAN equivalency in its usage**_. However, decoupling async helper from the big [original PR](https://github.com/nodejs/node/pull/11975) makes sense at this time for the following reasons:

* Allows us to carefully redesign and test async helper and not rush it in 
* Allows us to make forward progress on the [original PR](https://github.com/nodejs/node/pull/11975) without this being a blocker

We decided to **_temporarily_** take out async helper from the [original PR](https://github.com/nodejs/node/pull/11975) and commit to prioritizing the redesign work and open a new PR with the right design and validation done. (Hopefully, immediately after the original PR lands!)

kkoopa:
I see. I did not understand that it was merely temporary.

To clarify my second statement, I was trying to argue for keeping the libuv wrappers for the sake of encapsulation and consistency, but agree that it is not essential for the first implementation, especially given the short release window.

Provided that the experimental API makes it into node.js 8.0, what will its semver policy be from node core's side? Is it completely unaffected, partially unaffected, or wholly affected?

jasongin:
> Provided that the experimental API makes it into node.js 8.0, what will its semver policy be from node core's side?

Experimental features in node are explicitly not subject to semver policy, so they may change at any time. The header comment in `node_api.h` and the console output when enabling the feature via `--napi-modules` both warn about that.

In practice we will of course try to minimize breaking changes, as they are disruptive to anyone trying out the experimental feature. There are likely to be a few breaking changes early on, but stabilization is obviously a requirement for graduating out of experimental status. Much more common will be incremental addition of new APIs to expand coverage so that it is sufficient for a wider range of native addons. I don't think we've discussed yet whether we will try to align those kind of additions with semver-minor updates to node 8.x, even if we are technically not required to.

src/node_api.cc
```diff
@@ -2528,3 +2530,136 @@ napi_status napi_get_typedarray_info(napi_env env,
...
+  uvimpl::Work::Delete(reinterpret_cast<uvimpl::Work*>(work));
+
+  return napi_ok;
+}
+
+napi_status napi_queue_async_work(napi_env env, napi_async_work work) {
+  CHECK_ARG(env);
+  CHECK_ARG(work);
+
+  // Consider: Encapsulate the uv_loop_t into an opaque pointer parameter
```
jasongin, line 2642: If it's desirable we could add support for this later in a non-breaking way by adding another API, something like `napi_queue_work_on_loop()`. So I think this is fine for now.

jasongin (approved):

mhdawson (approved):
LGTM.  

I assume we just need to validate by bringing our ported modules up to date, and checking that all still works ok before we submit this to the core repo ?  If so that should probably be on the top of our priority list.

jasongin:
> we need to validate by bringing our ported modules up to date

Yes. And doing that will require an updated C++ wrapper that uses these new APIs. I'll prepare those changes in a branch.

sampsongao:
It seems like we got some trailing white spaces.

sampsongao:
```
node_api.cc:(.text+0x93e7): undefined reference to 'node::Environment::GetCurrent(v8::Isolate*)'
node_api.cc:(.text+0x93ef): undefined reference to 'node::Environment::event_loop() const'
```
Seems like we need to include env-inl.h

jasongin:
Updates here: https://github.com/nodejs/node-api/pull/10, tested with the NAPI node-canvas port.

jasongin:
@sampsongao I didn't see any build errors when I built with this PR on Windows. Maybe it is compiler-dependent? It does look like `evn-inl.h` should be required, if `Environment::event_loop()` is being inlined.

sampsongao:
After suggested changes, nanomsg is running find with the new async api.

src/node_api.cc
```diff
@@ -16,6 +16,7 @@
 #include <cmath>
 #include <vector>
 #include "node_api.h"
+#include "env.h"
```
sampsongao, line 20: #include "env-inl.h"

sampsongao (changes requested):

mhdawson:
Landed in main repo closing.

Issue #198, opened 2017-03-24
Last exception should be per-isolate


jasongin:
Currently in the N-API V8 implementation, the last exception is stored as a global. But that won't be correct when Node supports multiple isolates in the process. Then the last exception should be per-isolate.

addaleax:
Fwiw, the question of whether Node will eventually support multiple isolates out of the box hasn’t really been settled yet, but I suspect this would be cool to have for embedders like Electron.

gabrielschulhof:
I think if the last exception is per-isolate then so should `napi_get_last_error_info()` be - that is, it should accept the `napi_env` as its first argument. Thus, we need to store the last error in the `napi_env`. In turn, we should probably tack a structure onto the Isolate using [Isolate::SetData](https://v8docs.nodesource.com/node-7.4/d5/dda/classv8_1_1_isolate.html#a7acadfe7965997e9c386a05f098fbe36), but that does not let you provide a deleter for when the Isolate disappears. I guess it would be up to that which creates new isolates to also notify of their destruction, or up to V8 to provide such a callback in the future.

mhdawson:
Yes, every napi call should take a napi_env so that we always have that flexibility.  Looks like napi_get_last_error_info is the only one where we missed that.

mhdawson:
@gabrielschulhof is submitting a PR to add napi_env as a paratmeter to nap_get_last_error_info() something you can add to your list ?  That is something I think we want in before the PR lands in the core repo.

digitalinfinity:
@gabrielschulhof I think this is a great PR- however, my suggestion is to break it up into 2 PRs. The first one would change the signature of napi_get_last_error_info, but otherwise leave things as is. The second one would contain the rest of the implementation from this PR. My vote is to land the first one in the PR opened against nodejs/master currently, and land the second one after that PR lands, just to make the review of the larger PR smoother. Thoughts?

addaleax:
> just to make the review of the larger PR smoother

Fwiw so far it looks like I’m the only one actively reviewing the nodejs/master PR, and I don’t mind reviewing #206 as a single chunk (I’ll try to keep on top of the discussions here, too)

digitalinfinity:
@addaleax ok! Was trying to not make your life harder 😄

PR #206, opened 2017-03-28
napi: create napi_env as a real structure tied to the isolate

gabrielschulhof:
1. We define struct napi_env__ to include the isolate, the last
exception, and the info about the last error.

2. We instantiate one struct napi_env__ during module registration, but
only if there isn't one already attached to the isolate, perhaps from
the registration of a pervious module.

3. the last exception and the info about the last error are stored in
the napi_env__ struct which is associated with the isolate via
v8::Isolate::SetData().

Fixes https://github.com/nodejs/abi-stable-node/issues/198

### DIFF
--- a/src/node_api.cc
+++ b/src/node_api.cc
@@ -8,7 +8,6 @@
  *
  ******************************************************************************/
 
-
 #include <node_buffer.h>
 #include <node_object_wrap.h>
 #include <string.h>
@@ -17,16 +16,28 @@
 #include <vector>
 #include "node_api.h"
 
+napi_status napi_set_last_error(napi_env env,
+                                napi_status error_code,
+                                uint32_t engine_error_code = 0,
+                                void* engine_reserved = nullptr);
+void napi_clear_last_error(napi_env env);
+
+struct napi_env__ {
+  v8::Isolate* isolate;
+  v8::Persistent<v8::Value> last_exception;
+  napi_extended_error_info last_error;
+};
+
 namespace v8impl {
 
 //=== Conversion between V8 Isolate and napi_env ==========================
 
 napi_env JsEnvFromV8Isolate(v8::Isolate* isolate) {
-  return reinterpret_cast<napi_env>(isolate);
+  return static_cast<napi_env>(isolate->GetData(0));
 }
 
 v8::Isolate* V8IsolateFromJsEnv(napi_env e) {
-  return reinterpret_cast<v8::Isolate*>(e);
+  return e->isolate;
 }
 
 class HandleScopeWrapper {
@@ -252,23 +263,18 @@ class Reference : private Finalizer {
 
 class TryCatch : public v8::TryCatch {
  public:
-  explicit TryCatch(v8::Isolate* isolate)
-      : v8::TryCatch(isolate), _isolate(isolate) {}
+  explicit TryCatch(napi_env env)
+      : v8::TryCatch(env->isolate), _env(env) {}
 
   ~TryCatch() {
     if (HasCaught()) {
-      _the_exception.Reset(_isolate, Exception());
+      _env->last_exception.Reset(_env->isolate, Exception());
     }
   }
 
-  static v8::Persistent<v8::Value>& LastException() { return _the_exception; }
-
  private:
-  static v8::Persistent<v8::Value> _the_exception;
-  v8::Isolate* _isolate;
+  napi_env _env;
 };
-
-v8::Persistent<v8::Value> TryCatch::_the_exception;
 
 //=== Function napi_callback wrapper =================================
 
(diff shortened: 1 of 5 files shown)


gabrielschulhof:
https://ci.nodejs.org/view/All/job/node-test-commit-linux-abi/102/

src/node_api.cc
```diff
@@ -17,16 +16,27 @@
...
+  napi_extended_error_info last_error;
+};
+
 namespace v8impl {
 
 //=== Conversion between V8 Isolate and napi_env ==========================
 
 napi_env JsEnvFromV8Isolate(v8::Isolate* isolate) {
-  return reinterpret_cast<napi_env>(isolate);
+  return static_cast<napi_env>(isolate->GetData(0));
```
jasongin: How can we be sure that nobody else will use slot 0 on the isolate?

src/node_api.cc
```diff
@@ -508,9 +514,22 @@ void napi_module_register_cb(v8::Local<v8::Object> exports,
...
                              v8::Local<v8::Context> context,
                              void* priv) {
+  v8::Isolate* isolate = context->GetIsolate();
   napi_module* mod = static_cast<napi_module*>(priv);
+
+  // Can we be absolutely sure that, if isolate->GetData() returns non-NULL, it
+  // means that the pointer returned is one of ours (i.e. a napi_env)? What if
+  // some other framework makes use of slot 0? There really seems to be a need
+  // for a more robust way of assigning arbitrary data to the isolate.
+  if (isolate->GetData(0) == nullptr) {
```
addaleax, line 540: See the comment removed in 6d16048da634eaf2254250262b9016ab69e1f57b, I don’t think `0` will be available unconditionally. You can probably just use `3` again, but in any case, this should be defined as a macro/constant somewhere.

You could probably use `node::Environment::GetCurrent(isolate)` to access the current `node::Environment`, which is a grab-bag of information for cases like these. The downside of that is that that would make the contents of this file quite a bit more Node-specific.

jasongin: Could we add a field to `node::Environment` instead of using a slot on the isolate? That seems safer. I don't think we are at all worried about this code being node-specific; it is a node API after all.

addaleax: > Could we add a field to `node::Environment` instead of using a slot on the isolate? That seems safer.

Sure. A small other downside is that the current `v8::Context` needs to have been created by Node (or told about the Environment via `Environment::AssignToContext()`), but other contexts wouldn’t be usable for Node code anyway.

boingoing:
Looks pretty good and I think we could stash additional persistents in ```struct napi_env``` for things like caching ```Symbol.hasInstance```.

src/node_api.cc
gabrielschulhof: I think we need to solve this without additions to node itself, because we need this to work on versions of node that have not been modified so we can ship N-API as a stand-alone package.

gabrielschulhof:
@jasongin @addaleax what about simply using a global `std::map<Isolate*, napi_env>` in src/node_api.cc for now? After all, it's almost certain that only one item will be added to that map, and there never will be thousands of items added to it.

gabrielschulhof:
Also, given @mhdawson's comment at https://github.com/nodejs/abi-stable-node/pull/204#discussion_r108539433 about not necessarily tying `napi_env` one-to-one to `Isolate*` I would say that having the `std::map` would be good, because different implementations could use different keys instead of `Isolate*`, but the bottom line is they all may have to keep track of and distinguish between multiple instances of `napi_env`.

For now, I think we can tie to `Isolate*` one-to-one, but if we implement the mapping using a `std::map` then we can change the mapping later on without modifications to node.

gabrielschulhof:
OTOH, can we have one `napi_env` per module? If so, I believe we might be able to get away without tacking anything on to the `Isolate*` or onto anything else. After all, there are only two places where control passes *into* N-API:

1. Module registration
2. The thunk

So, if we create the `napi_env` during module registration, it comes back to us during `napi_create_function()`, and we can store it in the `FunctionCallbackData` for all functions we create.

If we go this way, we may end up creating multiple `napi_env` objects with exactly the same content, especially since we know that for now there is only one `Isolate*`, but OTOH we get complete independence from any global structure and we require only the opportunity to add context to functions we expose to JS - which is indispensable anyway, and which V8 provides to us *with style* via `SetInternalField()`.

gabrielschulhof:
I've created an alternative implementation using the per-module `napi_env`:
https://github.com/nodejs/abi-stable-node/compare/f4bbfeef79a70ce2198552a07f8c36ca7e710fa6...gabrielschulhof:198-per-module-environment
The trade-off is that we are creating more `napi_env` objects than necessary because we are creating them per-module rather than per-whatever-it-is-in-the-environment-we-should-key-to, OTOH we are divorcing ourselves even further from node, because we encapsulate all that we need from node (the `Isolate*`) in one place and one place alone.

This assumes of course that, once you expose a function to JS, it will always be called with the same `Isolate*` and that if it is to be called with a different `Isolate*` then the module will be re-registered using that `Isolate*`. Even if not, though, we can simply re-initialize our exising `napi_env` before we call the `napi_callback`:
```C++
...
env->isolate = isolate;
cb(env, cbinfo_wrapper);
...

gabrielschulhof:
https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-commit-linux-abi/105/

gabrielschulhof:
https://ci.nodejs.org/view/All/job/node-test-commit-linux-abi/107/

gabrielschulhof:
Actually, since we can't really go with `isolate->GetData(<anything>)` and we can't modify node because we want this to work with versions of node we haven't modified, I think the only approach is to go with the per-module environment.

gabrielschulhof:
https://ci.nodejs.org/view/All/job/node-test-commit-linux-abi/108/

gabrielschulhof:
https://ci.nodejs.org/view/All/job/node-test-commit-linux-abi/110/

gabrielschulhof:
https://ci.nodejs.org/view/All/job/node-test-commit-linux-abi/113/

jasongin:
Can we close this PR? I think https://github.com/nodejs/abi-stable-node/pull/208 is a better solution.

PR #207, opened 2017-03-28
Refcount for v8impl::Reference should be unsigned

boingoing:
Add ```Reference::RefCount()``` and check the refcount value before trying to call ```Reference::Unref()``` to avoid underflowing the refcount. This allows us to continue returning an error from ```napi_reference_unref``` if it is called with a reference that already has refcount set to zero.

### DIFF
--- a/src/node_api.cc
+++ b/src/node_api.cc
@@ -144,7 +144,7 @@ class Reference : private Finalizer {
  private:
   Reference(v8::Isolate* isolate,
             v8::Local<v8::Value> value,
-            int initial_refcount,
+            uint32_t initial_refcount,
             bool delete_self,
             napi_finalize finalize_callback,
             void* finalize_data,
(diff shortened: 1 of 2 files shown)


src/node_api.cc
```diff
@@ -191,15 +191,15 @@ class Reference : private Finalizer {
...
     if (++_refcount == 1) {
       _persistent.ClearWeak();
     }
 
     return _refcount;
   }
 
-  int Unref() {
+  uint32_t Unref() {
     if (--_refcount == 0) {
```
jasongin: Maybe I'm being paranoid, but I think this helper class should ensure there is not an underflow:
```C++
if (_refcount > 0) {
  --_refcount;
  ...
```

jasongin (approved):

src/node_api.cc
boingoing: Yeah, I wrestled with this myself. Figured the compiler could avoid actually doing two checks of ```_refcount == 0``` if we did check it here, too, but wasn't a big deal either since we only call ```Reference::Unref``` from one place, anyway. I'll just add the extra check.

boingoing:
Passing CI:
https://ci.nodejs.org/job/node-test-commit-linux-abi/103/

PR #208, opened 2017-03-29
198 per module environment

gabrielschulhof:
This is an alternative to #206. It doesn't tack the `napi_env` onto any global node object, but rather creates a new one during module registration, and then passes it around, including into the context for all callbacks/accessors/finalizers.


gabrielschulhof:
https://ci.nodejs.org/view/All/job/node-test-commit-linux-abi/106/

src/node_api.cc
```diff
@@ -17,17 +16,19 @@
...
+                                uint32_t engine_error_code = 0,
+                                void* engine_reserved = nullptr);
+void napi_clear_last_error(napi_env env);
 
-napi_env JsEnvFromV8Isolate(v8::Isolate* isolate) {
-  return reinterpret_cast<napi_env>(isolate);
-}
+struct napi_env__ {
+  v8::Isolate* isolate;
+  v8::Persistent<v8::Value> last_exception;
```
kkoopa: While it should not be an issue for normal operations, I would still suggest either using a trait for `Persistent` with `kResetInDestructor` true, or add a destructor that calls `last_exception.Reset()`.

gabrielschulhof:
https://ci.nodejs.org/view/All/job/node-test-commit-linux-abi/109/

src/node_api.cc
gabrielschulhof: Will do although, since this structure is per-module, unless modules are ever unloaded, the destructor will never be called.

gabrielschulhof:
https://ci.nodejs.org/view/All/job/node-test-commit-linux-abi/111/

gabrielschulhof:
https://ci.nodejs.org/view/All/job/node-test-commit-linux-abi/112/

kkoopa:
Yes, that means it comes at no cost, apart from a couple bytes of code size. However, can modules be unloaded? Will it be possible? Adding the cleanup code guards against that and signals to the reader that this does not leak, without even needing a comment stating the same.

kkoopa:
One more thing, I see a lot of macro definitions where the formal parameters are not wrapped in parentheses.

gabrielschulhof:
https://ci.nodejs.org/view/All/job/node-test-commit-linux-abi/115/

src/node_api.cc
```diff
@@ -328,12 +330,18 @@ class CallbackWrapperBase : public CallbackWrapper {
         v8::Local<v8::External>::Cast(
             _cbdata->GetInternalField(kInternalFieldIndex))->Value());
     v8::Isolate* isolate = _cbinfo.GetIsolate();
-    cb(v8impl::JsEnvFromV8Isolate(isolate), cbinfo_wrapper);
+    napi_env env = reinterpret_cast<napi_env>(
+        v8::Local<v8::External>::Cast(
+            _cbdata->GetInternalField(kEnvIndex))->Value());
```
addaleax: A `static_cast` should work fine

src/node_api.cc
```diff
@@ -481,13 +492,17 @@ v8::Local<v8::Object> CreateAccessorCallbackData(napi_env env,
...
+  v8::Isolate* isolate = env->isolate;
   v8::Local<v8::Context> context = isolate->GetCurrentContext();
 
   v8::Local<v8::ObjectTemplate> otpl = v8::ObjectTemplate::New(isolate);
   otpl->SetInternalFieldCount(v8impl::kAccessorFieldCount);
   v8::Local<v8::Object> cbdata = otpl->NewInstance(context).ToLocalChecked();
 
+  cbdata->SetInternalField(
+      v8impl::kEnvIndex,
+      v8::External::New(isolate, reinterpret_cast<void*>(env)));
```
addaleax: Ditto for `static_cast`

addaleax (approved):

src/node_api.cc
gabrielschulhof: @addaleax we retrieve the callback four lines above this one (you need to expand to see more context). Should I change that one to a `static_cast` as well?

addaleax: @gabrielschulhof If it works, yes. Tbh, I just didn’t comment because I didn’t remember whether casting between `void*` and function pointers works with a `static_cast`. :smile:

gabrielschulhof: Oh, and above that, we retrieve `the napi_callback_info` the same way, with a `reinterpret_cast`. Should we be using `static_cast` in all places where we place values into and retrieve values out of internal fields?

src/node_api.cc
gabrielschulhof: Same question for the `reinterpret_cast`s in the expanded context below.

src/node_api.cc
addaleax: @gabrielschulhof Basically, yes, use a `static_cast` wherever that works :) It could well be that I missed some when looking over the nodejs/master PR

gabrielschulhof: OK. TBH I'm quite bewildered by all these casts, coming from a C background where there's a real dearth of them, shall we say 😄

gabrielschulhof: OK, after trying and being gently held by the hand by the compiler, it seems only the `env` can be passed around via a `static_cast` and not any of the other internal fields.

kkoopa:
It does not work. void * is a data pointer, which is not the same as a function pointer in C++, however the POSIX standard requires them to be equivalent. The distinction is mostly a legacy from the times of segmented memory with long and short pointers, which you would only find in embedded systems nowadays.

gabrielschulhof:
@kkoopa wow! OK.

src/node_api.cc
```diff
@@ -469,6 +477,9 @@ v8::Local<v8::Object> CreateFunctionCallbackData(napi_env env,
       v8impl::kFunctionIndex,
       v8::External::New(isolate, reinterpret_cast<void*>(cb)));
   cbdata->SetInternalField(
+      v8impl::kEnvIndex,
+      v8::External::New(isolate, static_cast<void*>(env)));
```
jasongin: Nit: Move this up to before setting the function, just so things are in a consistent order.

jasongin (approved):

gabrielschulhof:
https://ci.nodejs.org/view/All/job/node-test-commit-linux-abi/116/

gabrielschulhof:
https://ci.nodejs.org/view/All/job/node-test-commit-linux-abi/123/

src/node_api.cc
```diff
@@ -629,38 +647,38 @@ const char* error_messages[] = {nullptr,
...
-napi_status napi_set_last_error(napi_status error_code,
-                                uint32_t engine_error_code = 0,
-                                void* engine_reserved = nullptr) {
-  static_last_error.error_code = error_code;
-  static_last_error.engine_error_code = engine_error_code;
-  static_last_error.engine_reserved = engine_reserved;
+napi_status napi_set_last_error(napi_env env, napi_status error_code,
+                                uint32_t engine_error_code,
+                                void* engine_reserved) {
+  env->last_error.error_code = error_code;
```
jasongin: This needs to check if `env` is null and then skip these lines, otherwise `CHECK_ARG(env, env)` will crash when `env` is null. At least it should be able to return `napi_invalid_arg`, even if it can't set the last error info in that case.

gabrielschulhof: Oh, right. Good point!

jasongin (changes requested):
Waiting for guard against null `env`.

gabrielschulhof:
@jasongin if somebody calls `napi_set_last_error()` with a `nullptr` `env` but a non-`napi_ok` error, then should I return `napi_invalid_arg` or the original error? That's a pretty twisted case, I must admit, and my gut tells me I should return the error passed in, because that's more relevant.

gabrielschulhof:
I guess `napi_set_last_error()` has to be able to work without an `env`, in which case all it does is return the passed-in error code, because the lack of an `env` could be the very error for which it was called.

jasongin:
> then should I return `napi_invalid_arg` or the original error?

Currently those two must be the same, because calling any API with a null `env` will immediately call `napi_set_last_error()` with `napi_invalid_arg`. But yeah, returning the original error seems better. (Remember `napi_set_last_error()` is not a public API.)

src/node_api.cc
```diff
@@ -658,9 +658,16 @@ void napi_clear_last_error(napi_env env) {
 napi_status napi_set_last_error(napi_env env, napi_status error_code,
                                 uint32_t engine_error_code,
                                 void* engine_reserved) {
-  env->last_error.error_code = error_code;
-  env->last_error.engine_error_code = engine_error_code;
-  env->last_error.engine_reserved = engine_reserved;
+
+  if (env == nullptr) {
+    if (error_code == napi_ok) {
```
jasongin: Is this really needed? We don't (and shouldn't) ever call `napi_set_last_error()` with `napi_ok`.

gabrielschulhof: Right, I guess we have `napi_clear_last_error()` for that. But wait a sec ... what do we do in `napi_clear_last_error()` if the `env` is `nullptr`?

gabrielschulhof: Perhaps `napi_clear_last_error()` needs to also return a `napi_status`.

jasongin: It should not be possible to reach `napi_clear_last_error()` with a null `env` because there is always a `CHECK_ARG(env, env)` before that.

jasongin: I don't think `napi_clear_last_error()` needs to return `napi_status` because it's not a public API and checks of `env` before it should ensure it can't fail.

gabrielschulhof: OK, I think we can safely assume that `env` is not `nullptr` whenever it's passed into a non-public API, because it's the public API's job to make sure it's not `nullptr`.

gabrielschulhof: ... except `napi_set_last_error()`.

gabrielschulhof: As we've dropped `NAPI_PREAMBLE(env)` is some places, we've also dropped the preliminary `CHECK_ARG(env, env)` that was a part of it. I'm putting those back in.

src/node_api.cc
```diff
@@ -629,38 +647,45 @@ const char* error_messages[] = {nullptr,
...
-                                void* engine_reserved = nullptr) {
-  static_last_error.error_code = error_code;
-  static_last_error.engine_error_code = engine_error_code;
-  static_last_error.engine_reserved = engine_reserved;
+napi_status napi_set_last_error(napi_env env, napi_status error_code,
+                                uint32_t engine_error_code,
+                                void* engine_reserved) {
+
+  if (env == nullptr) {
+    if (error_code == napi_ok) {
```
mhdawson: I'd agree that we should never be calling this with napi_ok, and it is an internal method so that would be a bug on our side.    Maybe an assert instead ?

src/node_api.cc
gabrielschulhof: Actually, those places were not accessing `env` in any way except as a whole, and passing it to functions/macros capable of dealing with a `nullptr` `env`, so it was actually OK.

gabrielschulhof: There's actually a superfluous `CHECK_ARG(env, env)` in `napi_get_value_double()` because that function only uses `env` to pass it to `RETURN_STATUS_IF_FALSE()` and nothing else, and `RETURN_STATUS_IF_FALSE()` is capable of dealing with a `nullptr` `env`. ... but I'll leave that in.

mhdawson:
napi_set_last_error() is not public, only used internally so its only our code that should be calling it.  As long as we are sure we always validate the env first we should be ok.

src/node_api.cc
mhdawson: Its better to leave in so that we don't have the case were code can get away without passing it in today, but then later fails if something changes and we do need to use it internally.

gabrielschulhof:
Perhaps we should react more harshly if `env` is ever `nullptr` - we should abort.

gabrielschulhof:
... because right now we null-check `env` the same way we null-check other arguments,
```C++
CHECK_ARG(env, env);
```
which ultimately calls `napi_set_last_error()` if the second parameter is `nullptr`.

jasongin (approved):

mhdawson (approved):
Overall LGTM

mhdawson:
In respect to the comment about aborting, it's not that friendly, but it may be better than having cases where the module developer does not pass in an env and also does not check the error code as that will almost certainly lead to unexpected problems.

gabrielschulhof:
I think that's the best way. We should take the approach that we cannot function without `env`, so the way I've set it up is this: all our public APIs start with `NAPI_PREAMBLE(env)` which in turn starts with `CHECK_ENV((env))`, which evaluates to `assert((env) != nullptr)` or, if try-catching is unnecessary, they start with `CHECK_ENV(env)`.

gabrielschulhof:
https://ci.nodejs.org/view/All/job/node-test-commit-linux-abi/126/

gabrielschulhof:
@addaleax, @jasongin, @mhdawson could you please take another look to make sure all is well with the new, more *assert*ive approach?

src/node_api.cc
```diff
@@ -557,65 +576,67 @@ void napi_module_register(napi_module* mod) {
...
   do {                                                                  \
     if (!(condition)) {                                                 \
-      return napi_set_last_error((status));                             \
+      return napi_set_last_error((env), (status));                      \
     }                                                                   \
   } while (0)
 
-#define CHECK_ARG(arg) RETURN_STATUS_IF_FALSE((arg), napi_invalid_arg)
+#define CHECK_ENV(env) \
+  assert((env) != nullptr)
```
jasongin: `assert()` does nothing in a release build. Instead, use something like: `if (!condition) node::FatalError(...)`

jasongin:
Personally I think returning an error code is sufficient and preferable to aborting. There is no need to bring down the whole process because there is a bug in some part of some native module's code. Aborts should be reserved for conditions when the *node engine or JS VM* cannot continue to execute reliably.

When using the C++ wrapper the return codes are always checked and will propagate up as a JavaScript exception if unhandled.

src/node_api.cc
```diff
@@ -328,12 +329,18 @@ class CallbackWrapperBase : public CallbackWrapper {
         v8::Local<v8::External>::Cast(
             _cbdata->GetInternalField(kInternalFieldIndex))->Value());
     v8::Isolate* isolate = _cbinfo.GetIsolate();
-    cb(v8impl::JsEnvFromV8Isolate(isolate), cbinfo_wrapper);
+    napi_env env = static_cast<napi_env>(
+        v8::Local<v8::External>::Cast(
+            _cbdata->GetInternalField(kEnvIndex))->Value());
+
+    // Make sure any errors encountered last time we were in N-API are gone.
+    napi_clear_last_error(env);
```
addaleax, line 337: Do you have an example of when this would be hit?

src/node_api.cc
addaleax: There’s also Node’s `CHECK` macros in `util.h` if you’re okay with including that.

jasongin: Node's `CHECK` macro also does nothing in a release build. But we do need this check enabled in a release build.

jasongin: Yes, that ends up calling `node::Assert()`, that then calls `Abort()`. But I'm still not convinced that aborting in this case is desirable.

addaleax: I’d say it’s nicer than segfaulting, and when it’s hit it’s always a sign that the program is in an invalid state… that sounds good enough for aborting to me?

jasongin: Certainly it's nicer than segfaulting. But I was suggesting returning `napi_invalid_arg` instead, as in @gabrielschulhof's previous revision of this PR. As explained in my other comment:

> There is no need to bring down the whole process because there is a bug in some part of some native module's code. Aborts should be reserved for conditions when the node engine or JS VM cannot continue to execute reliably.

> When using the C++ wrapper the return codes are always checked and will propagate up as a JavaScript exception if unhandled.

jasongin: It doesn't necessarily mean that even the native module is in an invalid state (at least not any more than any other NAPI error does). It could be possible for other calls to the native module to succeed.

addaleax: ¯\\\_(ツ)_/¯

I’m good either way.

src/node_api.cc
gabrielschulhof: I'll try to come up with an example.

src/node_api.cc
gabrielschulhof: > Aborts should be reserved for conditions when the node engine or JS VM cannot continue to execute reliably.

In this case it can't continue reliably, because no method of ours will work without an `env`. They'll all return `napi_invalid_arg`, so we might as well nip it in the bud.

jasongin: That particular call to the NAPI function can't continue with a null `env`. But that doesn't mean other calls to other NAPI functions in the same callback or other callbacks in the same native module, or other modules can't continue to work.

Why should this invalid argument error be treated any differently from any other invalid argument error? An invalid argument almost always indicates a bug in the code, but aborting the process at any sign of a bug _in an external module_ seems extreme to me.

src/node_api.cc
gabrielschulhof: I'll add a test for https://gist.github.com/gabrielschulhof/a01fa89e98ce1042dd6ec6776eada3b4

test/addons-napi/test_napi_status/test_napi_status.cc
```diff
@@ -0,0 +1,44 @@
...
+void createNapiError(napi_env env, napi_callback_info info) {
+  napi_status status;
+  napi_value value;
+  double double_value;
+
+  status = napi_create_string_utf8(env, "xyz", 3, &value);
+  if (status != napi_ok) return;
+
+  napi_get_value_double(env, value, &double_value);
+}
```
addaleax: Maybe add a check here that `napi_get_value_double` actually fails?

test/addons-napi/test_napi_status/test_napi_status.cc
```diff
@@ -0,0 +1,44 @@
...
+
+  napi_set_return_value(env, info, result);
+}
+
+void Init(napi_env env, napi_value exports, napi_value module, void* priv) {
+  napi_status status;
+
+  napi_property_descriptor descriptors[] = {
+      DECLARE_NAPI_METHOD(createNapiError),
+	  DECLARE_NAPI_METHOD(testNapiErrorCleanup)
```
addaleax: The indentation is off here ;)

addaleax (approved):

gabrielschulhof:
@jasongin we need to abort when
1. We have `NAPI_PREAMBLE()` because it may need to store the last exception, which is stored in the env.
2. We have `NAPI_ARG(env, <anything>)` because it needs to set the last error if `<anything>` is null, and the last error is stored in the env.

Really, the only time we really wouldn't need the environment is when a function did not need anything stored inside it, and always returned `napi_ok`. The closest function we have to that is `napi_typeof()`, but since it needs to make sure that `result` is not null, it has a `NAPI_ARG()` for `result`, which needs the env.

I've gone through the entire src/node_api.cc file and for all public APIs we have one of the above two conditions. So, I think we need to keep the abort, unless we are prepared to accept that `napi_get_last_error_info()` will sometimes not return the last error, because when the last error happened, then `env` was not passed in, so it couldn't be recorded, and thus, if `env` is around, after all, just not in that one call, then what you get from `napi_get_last_error_info()` is the second-last error, not the last error.

gabrielschulhof:
@jasongin if people are more likely to use the C++ wrapper anyway, since it looks a lot like NAN, then we need not worry about the abort so much, because the wrapper will always correctly pass in the env and the users of the wrapper need not worry about it.

jasongin:
> unless we are prepared to accept that napi_get_last_error_info() will sometimes not return the last error, because when the last error happened, then env was not passed in, so it couldn't be recorded, and thus, if env is around, after all, just not in that one call, then what you get from napi_get_last_error_info() is the second-last error, not the last error.

That's actually not a problem: with this change, `napi_get_last_error_info()` is defined as returning the last error _for the `env` that you give it_. So if you give it a null `env`, then it should return `napi_invalid_arg`, which would be the same as the last error for any other call that was attempted using a null `env`. There is no inconsistency.

So I still don't think aborting is the best response to a null `env`. But this is such a minor thing that hardly anyone will ever encounter, that I don't feel strongly about it.

PR #211, opened 2017-03-29
napi: add napi_env to napi_get_last_error_info()

gabrielschulhof:
This makes it necessary to pass napi_env to napi_get_last_error_info()
thereby ensuring that errors can be tied to the napi_env in the future.

Re https://github.com/nodejs/abi-stable-node/issues/198

### DIFF
--- a/src/node_api.cc
+++ b/src/node_api.cc
@@ -637,7 +637,8 @@ void napi_clear_last_error() {
   static_last_error.engine_reserved = nullptr;
 }
 
-const napi_extended_error_info* napi_get_last_error_info() {
+const napi_extended_error_info* napi_get_last_error_info(napi_env env) {
+  (void)env;
   static_assert(node::arraysize(error_messages) == napi_status_last,
       "Count of error messages must match count of error values");
   assert(static_last_error.error_code < napi_status_last);
--- a/src/node_api.cc
+++ b/src/node_api.cc
@@ -637,8 +637,20 @@ void napi_clear_last_error() {
   static_last_error.engine_reserved = nullptr;
 }
 
-const napi_extended_error_info* napi_get_last_error_info(napi_env env) {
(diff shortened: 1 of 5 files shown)


src/node_api.cc
```diff
@@ -637,7 +637,8 @@ void napi_clear_last_error() {
   static_last_error.engine_reserved = nullptr;
 }
 
-const napi_extended_error_info* napi_get_last_error_info() {
+const napi_extended_error_info* napi_get_last_error_info(napi_env env) {
+  (void)env;
```
boingoing: Assume this is to suppress unused formal argument warnings? Suppose what we want is actually to use something like ```CHECK_ARG(env);``` instead. That would require ```napi_get_last_error_info()``` to return ```napi_status```, though...

boingoing (approved):
Looks fine but should ```napi_get_last_error_info``` return a ```napi_status``` ?

jasongin:
> should napi_get_last_error_info return a napi_status

I'm leaning toward yes, both for consistency with other APIs and so that it can return an error code if you pass it a null/invalid `env`. The alternative is to indicate an error via a null return value, but I don't like that as much.

jasongin:
You'll need to rebase and merge after my last change. Sorry.

jasongin (changes requested):
Waiting for a change to return napi_status, or consensus that it is not needed.

jasongin:
I rebased and pushed a commit to @gabrielschulhof's branch.

jasongin:
CI passed: https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-commit-linux-abi/120/

boingoing (approved):
👍

gabrielschulhof:
@jasongin @boingoing thanks for jumping in!

PR #214, opened 2017-03-31
209 cache has instance

gabrielschulhof:
This PR depends on #208, so only the last commit is relevant. The approach is to try and retrieve `Symbol.hasInstance` the first time around. If that succeeds, store it in `env` in a `Persistent` and skip the retrieval the next time around. Otherwise set a boolean flag in `env` indicating that next time around just use the old-fashioned walk-up-the-prototype-chain implementation.


src/node_api.cc
```diff
@@ -23,12 +23,16 @@ void napi_clear_last_error(napi_env env);
...
+  explicit napi_env__(v8::Isolate* _isolate): isolate(_isolate),
+      has_instance_available(true), last_error() {}
   ~napi_env__() {
     last_exception.Reset();
+    has_instance.Reset();
   }
   v8::Isolate* isolate;
   v8::Persistent<v8::Value> last_exception;
+  v8::Persistent<v8::Value> has_instance;
+  bool has_instance_available;
```
jasongin: You can use `v8::PersistentBase::IsEmpty()` instead of keeping a separate boolean flag.

jasongin:
Are we sure this change is worthwhile? Could you benchmark `tests/addons-napi/test_instanceof` before and after the change to confirm that it yields a significant improvement?

gabrielschulhof:
These are the results when running `../../../node ./test.js` from test/addons-napi/test_instanceof with some `clock_gettime()` calls added to `napi_instanceof` (see links for where exactly I put the calls):

| nanoseconds	| [no caching](https://github.com/gabrielschulhof/abi-stable-node/commit/209-cache-has-instance-before-with-time)	| [caching (first time ignored)](https://github.com/gabrielschulhof/abi-stable-node/commit/209-cache-has-instance-after-with-time)	| [old-fashioned](https://github.com/gabrielschulhof/abi-stable-node/commit/209-cache-has-instance-old-fashioned) |
| --- | --- | --- | --- |		
|    |	(two five-digit outliers removed) |   |   |
|average:	|2735.183305	|1362.510309	|1597.163212|
|count:	|1162	|1164	|1158|

jasongin:
Great, thanks. So the caching does yield a big improvement. (Without being familiar with the performance characteristics of all the V8 APIs it was hard to be sure by looking at the code.)

jasongin (approved):

mhdawson:
landed in main repo closing.

Issue #153, opened 2017-03-16
NAPI Documentation


aruneshchandra:
- [x] Getting Started - Hitesh/Michael 
- [x] Data Types and Values (convert to and from C) - Hitesh
- [x] Abstract Operations (return napi objects or tell you the type) - Hitesh
- [x] Functions to call into JS - Hitesh
- [x] Functions to call back from JS  - Hitesh
- [x] Error Handling - Michael 
- [x] Memory Management - Michael 
- [x] Async APIs - Michael 
- [x] Module Registration  - Michael 
- [x] ObjectWrap - Jason

digitalinfinity:
Related C++ issue: https://github.com/nodejs/node-api/issues/6

jasongin:
Maybe there should be another section about **Properties**: getting, setting, and defining them.

digitalinfinity:
Added and self-assigned for now 😄

digitalinfinity:
I had a bunch of these docs sitting on my machine for a couple weeks- dusted them off, updated with the API changes since then and opened PRs #215, #216, #217, #218 and #219

mhdawson:
Will try to add my 2 cents tomorrow

aruneshchandra:
This is done and live
https://nodejs.org/dist/latest-v7.x/docs/api/n-api.html

PR #215, opened 2017-04-03
Add doc for N-API basic datatypes

digitalinfinity:
First stab at documentation for basic N-API datatypes and concepts. This could probably use some more examples, so would love feedback on what content is appropriate for this doc.

Ref: #153

### DIFF
--- a/napi_basics.md
+++ b/napi_basics.md
@@ -1,0 +1,120 @@
+# N-API - Basics
+
+N-API is an API exposed by Node to developers that allow them to author Native add-ons 
+in an ABI stable manner. This means that these add-ons can work against future versions
+of Node without recompilation, and developers get the additional benefit that their 
+add-ons can work independent of the underlying JavaScript Virtual Machine being used 
+by Node.
+
+APIs exposed in the N-API surface are generally used to create and manipulate JavaScript
+objects. Concepts and operations generally map to ideas specified in the ECMA262 Language
+SPecification. The APIs generally have the following properties:
+- All N-API APIs generally return a status code of type `napi_status`. This status 
+can be used to determine whether the API call succeeded or failed.
+- The actual value being returned by the API is returned in an out parameter.
+- All JavaScript objects are abstracted behind an opaque type named `napi_value`.
+- In case of an error status code, additional information can be obtained using 
+`napi_get_last_error_info`
+
+## N-API types
+
+N-API exposes the following fundamental datatypes as abstractions that are consumed by 
+the various APIs. These APIs should be treated as opaque types that can be introspected
+only with other N-API APIs.
+
+### *napi_env*
+
+`napi_env` is used to represent a context that the underlying N-API implementation can use 
+to persist VM specific state.
+<< REVIEW: What else can we say here? >>
+
+### *napi_value*
+This is an opaque pointer that is used to represent a JavaScript object.
+
+### *napi_ref*
+This is the abstraction to use to reference a `napi_value`. This allows for users 
+to manage the lifetimes of JavaScript objects, including defining their minimum
+lifetimes explicitly, and also being notified when they are garbage-collected.
+
+### *napi_handle_scope*
+This is an abstraction used to provide hints to the garbage collector about the location of 
+handles to GC pointers. napi_handle_scope allows the user to provide a variable stack
+separate from the thread stack, in a manner that is efficient for scan in case 
+the underlying VM uses a garbage collector with precise-semantics.
+N-API values are created within the context of a handle scope- if the user does not 
+create a handle scope, they are created within the default handle scope.
+Handle scopes are created using `napi_open_handle_scope` and are destroyed using 
+`napi_close_handle_scope`. Calling close can indicate to the GC that all napi_values 
+created during the lifetime of the handle scope are no longer referenced from the
+current stack frame.
+
+Consider creating a `napi_handle_scope` per function with a non-trivial amount of 
+objects created.
+
+### *napi_escapable_handle_scope*
+Escapable Handle Scopes are a special type of handle scope that are used to return
+values created within a particular handle scope to a parent scope.
+
+### *napi_callback_info*
+Opaque datatype that is passed to a callback function. It can be used for two purposes:
+- Get additional information about the context in which the callback was invoked
+- Set the return value of the callback
+
+### *napi_callback*
+Function pointer type for N-API APIs that call back into user-provided native code.
+Your callback function should satisfy the following signature:
+```
+typedef void (*napi_callback)(napi_env, napi_callback_info);
+```
+
+### *napi_finalize*
+Function pointer type for N-API APIs that allow the user to be notified when an object 
+is garbage collected. The user must provide a function satisfying the following signature
+which would get called upon the object's collection.
+```
+typedef void (*napi_finalize)(void* finalize_data, void* finalize_hint);
+```
+
+### *napi_status*
+Integral status code indicating the success or failure of a N-API API call.
+Currently, the following status codes are supported.
+```
+  napi_ok,
+  napi_invalid_arg,
+  napi_object_expected,
+  napi_string_expected,
+  napi_function_expected,
+  napi_number_expected,
+  napi_boolean_expected,
+  napi_generic_failure,
+  napi_pending_exception,
+  napi_status_last
+```
+If additional information is required upon an API returning a failed status, it can
+be obtained by calling `napi_get_last_error_info`.
+
+### *napi_extended_error_info*
+#### Definition
+```
+struct napi_extended_error_info {
+  const char* error_message;
+  void* engine_reserved;
+  uint32_t engine_error_code;
+  napi_status error_code;
+};
+```
+
+#### Members
+- error_message: UTF8-encoded string containing a VM-neutral description of the error
+- engine_reserved: Do not use.
+- engine_error_code: VM-specific error code
+- error_code: the N-API status code that originated with the current error
+
+
+
+
+
+
+
+
+
--- a/napi_basics.md
+++ b/napi_basics.md
@@ -1,16 +1,16 @@
 # N-API - Basics
 
-N-API is an API exposed by Node to developers that allow them to author Native 
-add-ons in an ABI stable manner. This means that these add-ons can work 
+N-API is an API exposed by Node to developers that allow them to author native 
+add-ons in an ABI-stable manner. This means that these add-ons can work 
 against future versions of Node without recompilation, and developers get the 
-additional benefit that their add-ons can work independent of the underlying 
+additional benefit that their add-ons can work independently of the underlying 
 JavaScript Virtual Machine being used by Node.
 
-APIs exposed in the N-API surface are generally used to create and manipulate
-JavaScript objects. Concepts and operations generally map to ideas specified 
+APIs exposed by the N-API surface are generally used to create and manipulate
+JavaScript values. Concepts and operations generally map to ideas specified 
 in the ECMA262 Language Specification. The APIs generally have the following 
 properties:
-- All N-API APIs generally return a status code of type `napi_status`. This
+- All N-API calls generally return a status code of type `napi_status`. This
   status can be used to determine whether the API call succeeded or failed.
 - The actual value being returned by the API is returned in an out parameter.
 - All JavaScript values are abstracted behind an opaque type named 
@@ -23,10 +23,10 @@ properties:
 
 N-API exposes the following fundamental datatypes as abstractions that are 
 consumed by the various APIs. These APIs should be treated as opaque types 
-that can be introspected only with other N-API APIs.
+that can be introspected only with other N-API calls.
 
 ### *napi_status*
-Integral status code indicating the success or failure of a N-API API call.
+Integral status code indicating the success or failure of a N-API call.
 Currently, the following status codes are supported.
 ```C
   napi_ok,
@@ -37,8 +37,7 @@ Currently, the following status codes are supported.
   napi_number_expected,
   napi_boolean_expected,
   napi_generic_failure,
-  napi_pending_exception,
-  napi_status_last
+  napi_pending_exception
 ```
 If additional information is required upon an API returning a failed status, 
 it can be obtained by calling `napi_get_last_error_info`.
@@ -46,12 +45,12 @@ it can be obtained by calling `napi_get_last_error_info`.
 ### *napi_extended_error_info*
 #### Definition
 ```C
-struct napi_extended_error_info {
+typedef struct {
   const char* error_message;
   void* engine_reserved;
   uint32_t engine_error_code;
   napi_status error_code;
-};
+} napi_extended_error_info;
 ```
 
 #### Members
@@ -59,8 +58,9 @@ struct napi_extended_error_info {
   the error
 - engine_reserved: Reserved for VM-specific error details. This is currently
   not implemented for any VM.
-- engine_error_code: VM-specific error code
-- error_code: the N-API status code that originated with the current error
+- engine_error_code: VM-specific error code. This is currently
+  not implemented for any VM.
+- error_code: the N-API status code that originated with the last error
 
 ### Additional Information
 
@@ -72,7 +72,7 @@ See the Error Handling section (<PLACEHOLDER>)
 implementation can use to persist VM-specific state. This structure is passed
 to native functions when they're invoked, and it must be passed back when 
 making N-API calls. Specifically, the same `napi_env` that was passed in when
-the initial native function was called should be passed to any subsequent 
+the initial native function was called must be passed to any subsequent 
 nested N-API calls. Caching the `napi_env` for the purpose of general reuse is
 not allowed.
 
@@ -90,26 +90,25 @@ create a new handle scope, N-API values will be created in the default handle
 scope. For any invocations of code outside the execution of a native method 
 (for instance, during a libuv callback invocation), the module is required to 
 create a scope before invoking any functions that can result in the creation
-of JavaScript objects.
+of JavaScript values.
 
 Handle scopes are created using `napi_open_handle_scope` and are destroyed 
 using `napi_close_handle_scope`. Calling close can indicate to the GC that all
 napi_values created during the lifetime of the handle scope are no longer 
 referenced from the current stack frame.
 
-Consider creating a `napi_handle_scope` per function if you create any handle 
-other than just the return value. Otherwise, the handle created will leak to 
-the parent handle scope which can lead to higher than expected (and necessary)
-memory usage.
+For more details, review the [Object Lifetime Management section](TODO://path).
 
 ### *napi_escapable_handle_scope*
-Escapable Handle Scopes are a special type of handle scope that are used to 
-return values created within a particular handle scope to a parent scope.
+Escapable handle Scopes are a special type of handle scope to return values
+created within a particular handle scope to a parent scope.
 
 ### *napi_ref*
 This is the abstraction to use to reference a `napi_value`. This allows for 
 users to manage the lifetimes of JavaScript values, including defining their 
 minimum lifetimes explicitly.
+
+For more details, review the [Object Lifetime Management section](TODO://path).
 
 ## N-API Callback types
 
@@ -121,19 +120,20 @@ purposes:
 - Set the return value of the callback
 
 ### *napi_callback*
-Function pointer type for N-API APIs that call back into user-provided native 
-code. Your callback function should satisfy the following signature:
+Function pointer type for user-provided native functions which are to be 
+exposed to JavaScript via N-API. Your callback function should satisfy the 
+following signature:
 ```C
 typedef void (*napi_callback)(napi_env, napi_callback_info);
 ```
 
 ### *napi_finalize*
-Function pointer type for N-API APIs that allows the user to be notified when 
-externally-owned data is ready to be cleaned up because the object that it 
-was associated with is garbage collected. The user must provide a function
-satisfying the following signature which would get called upon the object's
-collection. Currently, `napi_finalize` can be used to know when only objects
-with external data are collected.
+Function pointer type for add-on provided functions that allow the user to be
+notified when externally-owned data is ready to be cleaned up because the
+object that it was associated with is garbage-collected. The user must provide
+a function satisfying the following signature which would get called upon the
+object's collection. Currently, `napi_finalize` can be used for finding out when
+objects that have external data are collected.
 ```C
 typedef void (*napi_finalize)(void* finalize_data, void* finalize_hint);
 ```
--- a/napi_basics.md
+++ b/napi_basics.md
@@ -16,76 +16,14 @@ properties:
 - All JavaScript values are abstracted behind an opaque type named 
   `napi_value`.
 - In case of an error status code, additional information can be obtained 
-  using `napi_get_last_error_info`
+  using `napi_get_last_error_info`. More information can be found in the error
+  handling section (<PLACEHOLDER: link to error handling doc>)
 
-## N-API types
+## N-API General types
 
 N-API exposes the following fundamental datatypes as abstractions that are 
 consumed by the various APIs. These APIs should be treated as opaque types 
 that can be introspected only with other N-API APIs.
-
-### *napi_env*
-
-`napi_env` is used to represent a context that the underlying N-API 
-implementation can use to persist VM specific state.
-<< REVIEW: What else can we say here? >>
-
-### *napi_value*
-This is an opaque pointer that is used to represent a JavaScript value.
-
-### *napi_ref*
-This is the abstraction to use to reference a `napi_value`. This allows for 
-users to manage the lifetimes of JavaScript values, including defining their 
-minimum lifetimes explicitly, and also being notified when they are 
-garbage-collected.
-
-### *napi_handle_scope*
-This is an abstraction used to provide hints to the garbage collector about 
-the location of handles to GC pointers. napi_handle_scope allows the user to 
-provide a variable stack separate from the thread stack, in a manner that is 
-efficient for scan in case the underlying VM uses a garbage collector with 
-precise-semantics.
-N-API values are created within the context of a handle scope- if the user 
-does not create a handle scope, they are created within the default handle
-scope. In the case of a libuv callback, there is no default handle scope so 
-the user is required to supply one.
-Handle scopes are created using `napi_open_handle_scope` and are destroyed 
-using `napi_close_handle_scope`. Calling close can indicate to the GC that all
-napi_values created during the lifetime of the handle scope are no longer 
-referenced from the current stack frame.
-
-Consider creating a `napi_handle_scope` per function if you create any handle 
-other than just the return value. Otherwise, the handle created will leak to 
-the parent handle scope which can lead to higher than expected (and necessary)
-memory usage.
-
-### *napi_escapable_handle_scope*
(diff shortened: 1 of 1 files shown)


napi_basics.md
```diff
@@ -0,0 +1,120 @@
...
+to manage the lifetimes of JavaScript objects, including defining their minimum
+lifetimes explicitly, and also being notified when they are garbage-collected.
+
+### *napi_handle_scope*
+This is an abstraction used to provide hints to the garbage collector about the location of 
+handles to GC pointers. napi_handle_scope allows the user to provide a variable stack
+separate from the thread stack, in a manner that is efficient for scan in case 
+the underlying VM uses a garbage collector with precise-semantics.
+N-API values are created within the context of a handle scope- if the user does not 
+create a handle scope, they are created within the default handle scope.
```
kkoopa: There is no default (node-supplied) handle scope in libuv callbacks. For V8, the handle scope is non-optional there.

napi_basics.md
```diff
@@ -0,0 +1,120 @@
...
+Handle scopes are created using `napi_open_handle_scope` and are destroyed using 
+`napi_close_handle_scope`. Calling close can indicate to the GC that all napi_values 
+created during the lifetime of the handle scope are no longer referenced from the
+current stack frame.
+
+Consider creating a `napi_handle_scope` per function with a non-trivial amount of 
+objects created.
```
kkoopa: Unless the only handle created is the return value, there should be a handle scope. Else, they will leak to the surrounding handle scope, which might lead to higher memory usage than necessary.

napi_basics.md
```diff
@@ -0,0 +1,120 @@
...
+add-ons can work independent of the underlying JavaScript Virtual Machine being used 
+by Node.
+
+APIs exposed in the N-API surface are generally used to create and manipulate JavaScript
+objects. Concepts and operations generally map to ideas specified in the ECMA262 Language
+SPecification. The APIs generally have the following properties:
+- All N-API APIs generally return a status code of type `napi_status`. This status 
+can be used to determine whether the API call succeeded or failed.
+- The actual value being returned by the API is returned in an out parameter.
+- All JavaScript objects are abstracted behind an opaque type named `napi_value`.
```
jasongin: All JavaScript values (not just objects)

napi_basics.md
```diff
@@ -0,0 +1,120 @@
...
+only with other N-API APIs.
+
+### *napi_env*
+
+`napi_env` is used to represent a context that the underlying N-API implementation can use 
+to persist VM specific state.
+<< REVIEW: What else can we say here? >>
+
+### *napi_value*
+This is an opaque pointer that is used to represent a JavaScript object.
```
jasongin: JavaScript ~~object~~ value

napi_basics.md
```diff
@@ -0,0 +1,120 @@
...
+
+### *napi_ref*
+This is the abstraction to use to reference a `napi_value`. This allows for users 
+to manage the lifetimes of JavaScript objects, including defining their minimum
+lifetimes explicitly, and also being notified when they are garbage-collected.
```
jasongin: `napi_ref` does not actually expose finalize callbacks.

napi_basics.md
```diff
@@ -0,0 +1,120 @@
...
+`napi_env` is used to represent a context that the underlying N-API implementation can use 
+to persist VM specific state.
+<< REVIEW: What else can we say here? >>
+
+### *napi_value*
+This is an opaque pointer that is used to represent a JavaScript object.
+
+### *napi_ref*
+This is the abstraction to use to reference a `napi_value`. This allows for users 
+to manage the lifetimes of JavaScript objects, including defining their minimum
```
jasongin: JavaScript ~~objects~~ values

napi_basics.md
```diff
@@ -0,0 +1,120 @@
...
+  const char* error_message;
+  void* engine_reserved;
+  uint32_t engine_error_code;
+  napi_status error_code;
+};
+```
+
+#### Members
+- error_message: UTF8-encoded string containing a VM-neutral description of the error
+- engine_reserved: Do not use.
```
jasongin: I'd say "Reserved for VM-specific error details, currently not implemented for any VMs".

napi_basics.md
```diff
@@ -0,0 +1,120 @@
...
+### *napi_callback*
+Function pointer type for N-API APIs that call back into user-provided native code.
+Your callback function should satisfy the following signature:
+```
+typedef void (*napi_callback)(napi_env, napi_callback_info);
+```
+
+### *napi_finalize*
+Function pointer type for N-API APIs that allow the user to be notified when an object 
+is garbage collected. The user must provide a function satisfying the following signature
```
jasongin: ... allows the user to be notified _when externally-owned data is ready to be cleaned up_ because the object it was associated with is garbage-collected.

jasongin (changes requested):
Added some suggested corrections and clarifications.

napi_basics.md
digitalinfinity: Good point @kkoopa. I'll update the doc to call this case out- however, @jasongin and I were talking about this offline- we were wondering if this could be handled at the N-API layer or whether that would be too complex.

napi_basics.md
digitalinfinity: Good suggestion! Fixed

napi_basics.md
digitalinfinity: Did I misread the intent of `napi_wrap`? That returns a `napi_ref` and seems to take a finalize callback?
I suppose that's true if `napi_ref` was created only through `napi_create_reference`. However, you could determine if an object was being collected if you called `napi_wrap` on it (provided you actually had some external data on it I suppose)?

I'm torn on this one- I think this section could be reworded but I'm not sure exactly how. Thoughts?

digitalinfinity: To be more precise- I understand that the callback is called when the external data can be collected, however, that still implies that the object that was wrapped was collected

napi_basics.md
digitalinfinity: Fixed all instances

napi_basics.md
digitalinfinity: Changed

napi_basics.md
digitalinfinity: Fixed

digitalinfinity:
Thanks @jasongin. Pushed a commit that addresses your feedback

napi_basics.md
jasongin: All the APIs that wrap some kind of external data support finalize callbacks: `napi_wrap()`, `napi_create_external()`, `napi_create_external_buffer()`, `napi_create_external_arraybuffer()`. But the finalize callbacks are not necessarily associated with a reference, and it is possible to create a reference to an ordinary JavaScript value where there is no ability to have a finalize callback. (V8 does support weak-ref callbacks for any kind of JS values, but that is not exposed in N-API.)

I'd suggest removing the mention of the finalize callback from this section about `napi_ref`, and adding more detail to the section about the `napi_finalize` callback, describing how the callback is used specifically to clean up external data that is wrapped by JS values.

jasongin (changes requested):
Waiting for changes related to `napi_ref` and `napi_finalize`. I added a comment explaining more about those.

napi_basics.md
```diff
@@ -0,0 +1,120 @@
...
+
+APIs exposed in the N-API surface are generally used to create and manipulate JavaScript
+objects. Concepts and operations generally map to ideas specified in the ECMA262 Language
+SPecification. The APIs generally have the following properties:
+- All N-API APIs generally return a status code of type `napi_status`. This status 
+can be used to determine whether the API call succeeded or failed.
+- The actual value being returned by the API is returned in an out parameter.
+- All JavaScript values are abstracted behind an opaque type named `napi_value`.
+- In case of an error status code, additional information can be obtained using 
+`napi_get_last_error_info`
```
mhdawson: may add placeholder for reference to error handling section here ?

napi_basics.md
```diff
@@ -0,0 +1,120 @@
...
+## N-API types
+
+N-API exposes the following fundamental datatypes as abstractions that are consumed by 
+the various APIs. These APIs should be treated as opaque types that can be introspected
+only with other N-API APIs.
+
+### *napi_env*
+
+`napi_env` is used to represent a context that the underlying N-API implementation can use 
+to persist VM specific state.
```
mhdawson: We can mention that it is passed to native functions when they are invoked and that it must be passed back when making N-API calls.  In addition, the same napi_env that was passed when the initial native function was called should passed into any N-API calls (ie don't cache it once and then use everywhere).

napi_basics.md
```diff
@@ -0,0 +1,120 @@
...
+### *napi_ref*
+This is the abstraction to use to reference a `napi_value`. This allows for users 
+to manage the lifetimes of JavaScript values, including defining their minimum
+lifetimes explicitly, and also being notified when they are garbage-collected.
+
+### *napi_handle_scope*
+This is an abstraction used to provide hints to the garbage collector about the location of 
+handles to GC pointers. napi_handle_scope allows the user to provide a variable stack
+separate from the thread stack, in a manner that is efficient for scan in case 
+the underlying VM uses a garbage collector with precise-semantics.
```
mhdawson: This might be a little too precise.  We might want to limit explanation to how scopes control the object lifetime and how the application has to use scopes to modify the default lifespan.

napi_basics.md
```diff
@@ -0,0 +1,120 @@
...
+N-API values are created within the context of a handle scope- if the user does not 
+create a handle scope, they are created within the default handle scope. In the case of a 
+libuv callback, there is no default handle scope so the user is required to supply one.
+Handle scopes are created using `napi_open_handle_scope` and are destroyed using
```
mhdawson: We might want to generalize this to say that when a native method is called from JavaScript a default scope will exist.  For any invocations of code outside the execution of a native method (for example if using libuv directly) the application must create a scope before any actions that can result in JavaScript objects being created.

napi_basics.md
```diff
@@ -0,0 +1,120 @@
...
+  napi_function_expected,
+  napi_number_expected,
+  napi_boolean_expected,
+  napi_generic_failure,
+  napi_pending_exception,
+  napi_status_last
+```
+If additional information is required upon an API returning a failed status, it can
+be obtained by calling `napi_get_last_error_info`.
+
```
mhdawson: I'd be tempted to move napi_status and napi_extended_error_info to near the top under env, as something that applies to any code that will be written so that we have the most general concepts up front and the ones with more specific context later on.

It might also be useful to have subsections that capture where the types will be used something like:

General
Memory management
References

In the later ones we can include a reference to the more detailed sections with the functions that use those types.

mhdawson (changes requested):
Specific comments + 
- General comment that line lengths need to be <=80 chars.

napi_basics.md
digitalinfinity: Talked to @jasongin offline- the part I was missing was that napi_wrap cannot be called on arbitrary objects so there is no current mechanism in NAPI to be notified when an arbitrary napi_value is collected. Updating the doc.

napi_basics.md
digitalinfinity: Done

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Took the first suggestion. Tried to take the second, let me know if what I did was what you had in mind.

napi_basics.md
```diff
@@ -0,0 +1,139 @@
+# N-API - Basics
+
+N-API is an API exposed by Node to developers that allow them to author Native
```
gabrielschulhof: "... that *allows* them ..."
"Native" need not be capitalized.

napi_basics.md
```diff
@@ -0,0 +1,139 @@
+# N-API - Basics
+
+N-API is an API exposed by Node to developers that allow them to author Native 
+add-ons in an ABI stable manner. This means that these add-ons can work
```
gabrielschulhof: "... and ABI-stable manner."

napi_basics.md
```diff
@@ -0,0 +1,139 @@
+# N-API - Basics
+
+N-API is an API exposed by Node to developers that allow them to author Native 
+add-ons in an ABI stable manner. This means that these add-ons can work 
+against future versions of Node without recompilation, and developers get the 
+additional benefit that their add-ons can work independent of the underlying
```
gabrielschulhof: "... work independent*ly* of ..."

napi_basics.md
```diff
@@ -0,0 +1,139 @@
+# N-API - Basics
+
+N-API is an API exposed by Node to developers that allow them to author Native 
+add-ons in an ABI stable manner. This means that these add-ons can work 
+against future versions of Node without recompilation, and developers get the 
+additional benefit that their add-ons can work independent of the underlying 
+JavaScript Virtual Machine being used by Node.
+
+APIs exposed in the N-API surface are generally used to create and manipulate
```
gabrielschulhof: "... exposed *on or by* the N-API ..."

napi_basics.md
```diff
@@ -0,0 +1,139 @@
...
+JavaScript objects. Concepts and operations generally map to ideas specified
```
gabrielschulhof: "JavaScript *values*."

napi_basics.md
```diff
@@ -0,0 +1,139 @@
...
+add-ons in an ABI stable manner. This means that these add-ons can work 
+against future versions of Node without recompilation, and developers get the 
+additional benefit that their add-ons can work independent of the underlying 
+JavaScript Virtual Machine being used by Node.
+
+APIs exposed in the N-API surface are generally used to create and manipulate
+JavaScript objects. Concepts and operations generally map to ideas specified 
+in the ECMA262 Language Specification. The APIs generally have the following 
+properties:
+- All N-API APIs generally return a status code of type `napi_status`. This
```
gabrielschulhof: I would prefer "N-API calls" or "N-API members". Perhaps it's easiest to just say "All APIs generally ..." since the whole document is about N-API.

napi_basics.md
```diff
@@ -0,0 +1,139 @@
...
+  napi_ok,
+  napi_invalid_arg,
+  napi_object_expected,
+  napi_string_expected,
+  napi_function_expected,
+  napi_number_expected,
+  napi_boolean_expected,
+  napi_generic_failure,
+  napi_pending_exception,
+  napi_status_last
```
gabrielschulhof: This isn't really a status that is ever returned.

napi_basics.md
```diff
@@ -0,0 +1,139 @@
...
+```
+If additional information is required upon an API returning a failed status, 
+it can be obtained by calling `napi_get_last_error_info`.
+
+### *napi_extended_error_info*
+#### Definition
+```C
+struct napi_extended_error_info {
```
gabrielschulhof: This is now `typedef struct {...} napi_extended_error_info;` to accommodate a C compiler.

napi_basics.md
```diff
@@ -0,0 +1,139 @@
...
+  napi_status error_code;
+};
+```
+
+#### Members
+- error_message: UTF8-encoded string containing a VM-neutral description of 
+  the error
+- engine_reserved: Reserved for VM-specific error details. This is currently
+  not implemented for any VM.
+- engine_error_code: VM-specific error code
```
gabrielschulhof: This is also currently unimplemented, IIRC. If so, this should also have "This is currently not implemented for any VM."

napi_basics.md
```diff
@@ -0,0 +1,139 @@
...
+- error_code: the N-API status code that originated with the current error
```
gabrielschulhof: "... with the *last* error", right?

napi_basics.md
```diff
@@ -0,0 +1,139 @@
...
+
+See the Error Handling section (<PLACEHOLDER>)
+
+### *napi_env*
+
+`napi_env` is used to represent a context that the underlying N-API 
+implementation can use to persist VM-specific state. This structure is passed
+to native functions when they're invoked, and it must be passed back when 
+making N-API calls. Specifically, the same `napi_env` that was passed in when
+the initial native function was called should be passed to any subsequent
```
gabrielschulhof: I would say "... *must* be passed..."

napi_basics.md
```diff
@@ -0,0 +1,139 @@
...
+### *napi_value*
+This is an opaque pointer that is used to represent a JavaScript value.
+
+## N-API Memory Management types
+
+### *napi_handle_scope*
+This is an abstraction used to control and modify the lifetime of objects 
+created within a particular scope. In general, N-API values are created within
+the context of a handle scope. When a native method is called from 
+JavaScript, a default handle scope will exist- if the user does not explicitly
```
gabrielschulhof, line 88: "... will exist<i>. Thus,</i> if..."

napi_basics.md
```diff
@@ -0,0 +1,139 @@
...
+create a new handle scope, N-API values will be created in the default handle
+scope. For any invocations of code outside the execution of a native method 
+(for instance, during a libuv callback invocation), the module is required to 
+create a scope before invoking any functions that can result in the creation
+of JavaScript objects.
```
gabrielschulhof: *... JavaScript *values*."

napi_basics.md
```diff
@@ -0,0 +1,139 @@
...
+
+Handle scopes are created using `napi_open_handle_scope` and are destroyed 
+using `napi_close_handle_scope`. Calling close can indicate to the GC that all
+napi_values created during the lifetime of the handle scope are no longer 
+referenced from the current stack frame.
+
+Consider creating a `napi_handle_scope` per function if you create any handle 
+other than just the return value. Otherwise, the handle created will leak to 
+the parent handle scope which can lead to higher than expected (and necessary)
+memory usage.
```
gabrielschulhof: We already have a section about object lifetime management in https://github.com/nodejs/abi-stable-node/pull/171/files. This repeats some of that information.

napi_basics.md
```diff
@@ -0,0 +1,139 @@
...
+
+### *napi_escapable_handle_scope*
+Escapable Handle Scopes are a special type of handle scope that are used to
```
gabrielschulhof: "Escapable *handle scopes* are ..."

gabrielschulhof: "... of handle scope <strike>that are used </strike> to ..."

napi_basics.md
```diff
@@ -0,0 +1,139 @@
...
+memory usage.
+
+### *napi_escapable_handle_scope*
+Escapable Handle Scopes are a special type of handle scope that are used to 
+return values created within a particular handle scope to a parent scope.
+
+### *napi_ref*
+This is the abstraction to use to reference a `napi_value`. This allows for 
+users to manage the lifetimes of JavaScript values, including defining their 
+minimum lifetimes explicitly.
```
gabrielschulhof, line 109: This is also covered by https://github.com/nodejs/abi-stable-node/pull/171/files. Perhaps there should be a link to that document here.

napi_basics.md
```diff
@@ -0,0 +1,139 @@
...
+
+### *napi_callback_info*
+Opaque datatype that is passed to a callback function. It can be used for two
+purposes:
+- Get additional information about the context in which the callback was 
+  invoked
+- Set the return value of the callback
+
+### *napi_callback*
+Function pointer type for N-API APIs that call back into user-provided native
```
gabrielschulhof: Instead of the first sentence I would write "Function pointer type for user-provided native functions which are to be exposed to JavaScript via N-API."

napi_basics.md
```diff
@@ -0,0 +1,139 @@
...
+### *napi_callback*
+Function pointer type for N-API APIs that call back into user-provided native 
+code. Your callback function should satisfy the following signature:
+```C
+typedef void (*napi_callback)(napi_env, napi_callback_info);
+```
+
+### *napi_finalize*
+Function pointer type for N-API APIs that allows the user to be notified when 
+externally-owned data is ready to be cleaned up because the object that it
```
gabrielschulhof: "... because the object *with which it was associated* is garbage<b>-</b>collected." (hyphen between "garbage" and "collected")

napi_basics.md
```diff
@@ -0,0 +1,139 @@
...
+
+### *napi_callback*
+Function pointer type for N-API APIs that call back into user-provided native 
+code. Your callback function should satisfy the following signature:
+```C
+typedef void (*napi_callback)(napi_env, napi_callback_info);
+```
+
+### *napi_finalize*
+Function pointer type for N-API APIs that allows the user to be notified when
```
gabrielschulhof: I would write "Function pointer type for *add-on-provided functions that allow* the user ..."

napi_basics.md
```diff
@@ -0,0 +1,139 @@
...
+```C
+typedef void (*napi_callback)(napi_env, napi_callback_info);
+```
+
+### *napi_finalize*
+Function pointer type for N-API APIs that allows the user to be notified when 
+externally-owned data is ready to be cleaned up because the object that it 
+was associated with is garbage collected. The user must provide a function
+satisfying the following signature which would get called upon the object's
+collection. Currently, `napi_finalize` can be used to know when only objects
```
gabrielschulhof: I would write "Currently, &#96;napi_finalize&#96; can only be used for finding out when objects that have external data are collected."

gabrielschulhof (changes requested):

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

digitalinfinity:
Thanks @jasongin @mhdawson @gabrielschulhof - push a new commit to address feedback

jasongin (approved):

mhdawson (approved):
Changes look good to me LGTM.

napi_basics.md
```diff
@@ -0,0 +1,139 @@
...
+- All JavaScript values are abstracted behind an opaque type named 
+  `napi_value`.
+- In case of an error status code, additional information can be obtained 
+  using `napi_get_last_error_info`. More information can be found in the error
+  handling section (<PLACEHOLDER: link to error handling doc>)
+
+## N-API General types
+
+N-API exposes the following fundamental datatypes as abstractions that are 
+consumed by the various APIs. These APIs should be treated as opaque types
```
gabrielschulhof: "These types should be treated as opaque, introspectable only with other N-API calls."

napi_basics.md
```diff
@@ -0,0 +1,139 @@
...
+the context of a handle scope. When a native method is called from 
+JavaScript, a default handle scope will exist- if the user does not explicitly
+create a new handle scope, N-API values will be created in the default handle
+scope. For any invocations of code outside the execution of a native method 
+(for instance, during a libuv callback invocation), the module is required to 
+create a scope before invoking any functions that can result in the creation
+of JavaScript values.
+
+Handle scopes are created using `napi_open_handle_scope` and are destroyed 
+using `napi_close_handle_scope`. Calling close can indicate to the GC that all
```
gabrielschulhof: "...~~Calling close~~Closing the scope..."

napi_basics.md
```diff
@@ -0,0 +1,139 @@
...
+napi_values created during the lifetime of the handle scope are no longer
```
gabrielschulhof: &#96;napi_value&#96;s

napi_basics.md
```diff
@@ -0,0 +1,139 @@
...
+
+Handle scopes are created using `napi_open_handle_scope` and are destroyed 
+using `napi_close_handle_scope`. Calling close can indicate to the GC that all
+napi_values created during the lifetime of the handle scope are no longer 
+referenced from the current stack frame.
+
+For more details, review the [Object Lifetime Management section](TODO://path).
+
+### *napi_escapable_handle_scope*
+Escapable handle Scopes are a special type of handle scope to return values
```
gabrielschulhof: No need to capitalize "Scopes"

napi_basics.md
```diff
@@ -0,0 +1,139 @@
...
+exposed to JavaScript via N-API. Your callback function should satisfy the 
+following signature:
+```C
+typedef void (*napi_callback)(napi_env, napi_callback_info);
+```
+
+### *napi_finalize*
+Function pointer type for add-on provided functions that allow the user to be
+notified when externally-owned data is ready to be cleaned up because the
+object that it was associated with is garbage-collected. The user must provide
```
gabrielschulhof: "...~~that it was associated with~~with which it was associated..."

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

napi_basics.md
digitalinfinity: Fixed

digitalinfinity:
Thanks @gabrielschulhof - fixed the comments you raised. Please take another look.

mhdawson:
landed as fc3b6369c7d69fe97c7c56f028aaa7e168ec6d32

PR #222, opened 2017-04-04
Document Object Wrap APIs

jasongin:


napi_object_wrap.md
```diff
@@ -0,0 +1,128 @@
...
+   the `data' property of the callback info
+ - `[in]  property_count`: Number of items in the `properties` array argument
+ - `[in]  properties`: Array of property descriptors describing static and
+   instance data properties, accessors, and methods on the class
+   See `napi_property_descriptor`
+ - `[out] result`: A `napi_value` representing the constructor function for
+   the class
+
+#### Description
+Projects a C++ class to JavaScript. Specifically, the C++ constructor callback
```
mhdawson: The "Projects" terminology is not resonating with me.  Would something like "Binds" work ?

jasongin: Hmm... the term "binds" implies to me that a higher layer (JavaScript) is proactively adapting to connect downward to a lower layer (C++). But in this case the C++ layer is proactively making itself accessible to JavaScript.

kkoopa: Associates

jasongin: I pushed a commit where I revised some of the sections to add more detail and hopefully more clarity, and I avoided the term "projects".

mhdawson (approved):
I like the updated version LGTM.

PR #233, opened 2017-04-19
Update node-chakracore branch for N-API changes

jasongin:
- Copy `node_api*` headers from node master branch
 - Rename `node_jsrtapi.cc` to `node_api_jsrt.cc`
 - Update module registration APIs and logic in `node.cc`
 - Move async APIs implementation into `node_api_jsrt.cc`
 - Update N-API JSRT implementations for _all_ the recent API changes
 - Rename tests directory from `addons-abi` to `addons-napi`
 - Copy all `addons-napi` tests from node master branch

All the `addons-napi` tests (identical to those now in node master) build successfully, and most of them pass with the N-API JSRT implementation. There are 3 failures that will require some non-trivial work to address, so I opened issues to track them separately:

 1. `addons_napi/test_buffer` fails due to a finalizer callback not being invoked after an explicit call to `gc()`. #237
 2. `addons_napi/test_instanceof` fails because `Symbol.hasInstance` is not handled by either `JsInstanceOf()` or the JSRT `napi_instanceof()` implementation. #236
 3. `addons_napi/test_string` fails because there are no JSRT string APIs for Latin1 encoding, so the JSRT `napi_create_string_latin1()` and `napi_get_value_string_latin1()` incorrectly use UTF-8 encoding instead. #235 (Also note TODOs in the diff below.)

Sorry, this is a huge diff. Reviewers should skip over all the test files and headers, since those are just copied from the node master branch. The interesting changes are all in `node_api_jsrt.cc`.


src/node.cc
```diff
@@ -1,4 +1,4 @@
-#include "node.h"
+﻿#include "node.h"
```
MSLaguana: Is this a rogue BOM? can't see any non-whitespace changes here.

jasongin: Yes.. VS 2017 likes to insert those. I'll fix it.

src/node_api_jsrt.cc
```diff
@@ -176,6 +217,9 @@ namespace jsrtimpl {
       jsrtimpl::ExternalCallback* externalCallback =
         reinterpret_cast<jsrtimpl::ExternalCallback*>(callbackState);
 
+      // Make sure any errors encountered last time we were in N-API are gone.
```
digitalinfinity, line 220: Is this the behavior in v8 too? I'm not sure if we want the previous error to be silently swallowed here

src/node_api_jsrt.cc
```diff
@@ -221,43 +265,100 @@ namespace jsrtimpl {
...
 
-#define CHECK_JSRT(expr)                                                \
-  do {                                                                  \
-    JsErrorCode err = (expr);                                           \
-    if (err != JsNoError) return napi_set_last_error(err);              \
-  } while(0)
+      CHECK_JSRT(JsCreatePropertyIdUtf8(
+        reinterpret_cast<char*>(name.data()), length, propertyId));
+    } else {
+      CHECK_JSRT(JsGetPropertyIdFromSymbol(key, propertyId));
```
digitalinfinity, line 290: Should we not check explicitly here for keyType being a symbol? Should JSRT expose a symbol keytype to make this check easier here? Or are you just relying on JsGetPropertyIdFromSymbol to fail if anything other than a symbol is passed in?

src/node_api_jsrt.cc
jasongin: Yes, this is correct, and is consistent with NAPI-V8. The "last error" should only apply to the _last_ call to N-API, so any error from a prior call should be cleared.

src/node_api_jsrt.cc
jasongin: Yes, I'm expecting `JsGetPropertyIdFromSymbol()` to fail if the key is not a symbol.

src/node_api_jsrt.cc
```diff
@@ -221,43 +265,100 @@ namespace jsrtimpl {
...
-    napi_status status = (expr);                                        \
-    if (status != napi_ok) return status;                               \
-  } while(0)
+  inline napi_status
+  JsNameValueFromPropertyDescriptor(const napi_property_descriptor* p,
+                                    napi_value* name) {
+    if (p->utf8name != nullptr) {
+      CHECK_JSRT(JsCreateStringUtf8(
+        reinterpret_cast<const uint8_t*>(p->utf8name),
+        strlen(p->utf8name),
```
digitalinfinity, line 313: Do we need to deal with embedded nulls here?

src/node_api_jsrt.cc
```diff
@@ -382,131 +555,127 @@ napi_status napi_define_class(napi_env e,
...
         CHECK_JSRT(JsSetProperty(descriptor, setProperty, setter, true));
       }
-    }
-    else {
+    } else if (p->method != nullptr) {
+      napi_value property_name;
+      CHECK_NAPI(jsrtimpl::JsNameValueFromPropertyDescriptor(p, &property_name));
+
+      JsPropertyIdRef valueProperty;
+      CHECK_JSRT(JsCreatePropertyIdUtf8("value", 5, &valueProperty));
```
digitalinfinity, line 667: This is being created in a loop- in the chakrashim, there is a perf optimization to cache some of the more frequently used propertyidrefs- we should have a todo to do the same in napi-jsrt

digitalinfinity (approved):

src/node_api_jsrt.cc
addaleax: That’s not really possible without also having a way of specifying the length in the API – I think that’s okay, though.

jasongin: If you want to use something other than a null-terminated UTF-8 string as the property key, you have the option to create a `napi_value` first and assign that to the `napi_property_descriptor.name` field instead of using the `utf8name` field.

src/node_api_jsrt.cc
jasongin: Opened https://github.com/nodejs/abi-stable-node/issues/234

src/node.cc
jasongin: Fixed here and also in `node_api_jsrt.cc`

PR #277, opened 2017-10-03
doc: updated README.md

mhdawson:


mhdawson:
Landing as we need this to be updated before NI presentation tomorrow. Can fix up later if there are any comments/updates needed.

mhdawson:
Landed as df9d07cace361fbe94a1a021e38d562ec4b6c6c5

PR #388, opened 2019-10-22
Add title to all readme sections

NickNaso:
I provide the title to the sections in the readme file so they can be linked by external resources like slides other issues and slides.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,14 +21,14 @@ will be warned that it is experimental as follows:
 
 Node.js versions 8.12.0 and above provide N-API as a stable feature.
 
-**Branches**
+## **Branches**
 
 Currently this repo is being used only for meta issue management and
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,14 +21,14 @@ will be warned that it is experimental as follows:
 
 Node.js versions 8.12.0 and above provide N-API as a stable feature.
 
-**Branches**
+## **Branches**
```
legendecas:
```suggestion
## Branches
```

Then we don't need these asterisks right?

NickNaso: Yes I can remove the asterisks from every title because they are unnecessary. Do you agree?

legendecas: Definitely, yes.

mhdawson (approved):
LGTM

legendecas (approved):

PR #447, opened 2023-02-13
add emnapi to node-api-engine-bindings.md

toyobayashi:
https://github.com/nodejs/node-addon-api/pull/1283#issuecomment-1427637888

Not sure if it is the correct position to add. emnapi is not a "runtime", but something (a library?) that can help users write bindings between wasm and JavaScript in a same way of Node-API, itself need to be run on a JavaScript runtime such as browser or Node.js.

### DIFF
--- a/node-api-engine-bindings.md
+++ b/node-api-engine-bindings.md
@@ -67,6 +67,7 @@ language to implement native addons for Node.js through Node-API:
 |[electron](https://github.com/electron/electron)| C++|
 |[iotjs](https://github.com/jerryscript-project/iotjs)| C|
 |[veil](https://github.com/lightsourceengine/veil)| C|
+|[emnapi](https://github.com/toyobayashi/emnapi)| C / JavaScript|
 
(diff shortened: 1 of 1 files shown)


node-api-engine-bindings.md
```diff
@@ -67,6 +67,7 @@ language to implement native addons for Node.js through Node-API:
 |[electron](https://github.com/electron/electron)| C++|
 |[iotjs](https://github.com/jerryscript-project/iotjs)| C|
 |[veil](https://github.com/lightsourceengine/veil)| C|
+|[emnapi](https://github.com/toyobayashi/emnapi)| C / JavaScript|
```
legendecas: Would you mind sorting these entries alphabetically?

legendecas (approved):

node-api-engine-bindings.md
toyobayashi: done

mhdawson (approved):
LGTM

PR #453, opened 2023-05-30
doc: add Node-API v9 badge

KevinEady:
This depends on the release of https://github.com/nodejs/node/pull/48151 . 

This badge URL will be used in the README in https://github.com/nodejs/node-addon-api#badges .


legendecas (approved):

legendecas:
Just realized that we'll also need to add the badge to https://github.com/nodejs/abi-stable-node#badges. Would you mind adding it too?

